PR #173, opened 2022-06-28
Define `LangVersion` as "10.0" and only once

mrlacey:
Fixes CommunityToolkit/Tooling-Windows-Submodule#64
Also, removes some properties that were unnecessarily set more than once.

Note. This will not currently pass due to a bug in the version of Uno.UI (4.2.6) that is currently being used. Updating this to the latest stable version breaks other things.

Opening this now (as a draft) so that it's ready when related issues (#174) have been addressed.

This replaces CommunityToolkit/Labs-Windows#150 as that branch had become messy while exploring the cause of related issues.

### DIFF
--- a/Windows.Toolkit.Common.props
+++ b/Windows.Toolkit.Common.props
@@ -17,7 +17,7 @@
     <Features>Strict</Features>
     <Nullable>enable</Nullable>
     <WarningsAsErrors>nullable</WarningsAsErrors>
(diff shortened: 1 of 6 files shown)


Arlodotexe:
Moving to the submodule repo. We'll consolidate the `<LangVersion>` in a PR there. Additionally, we'll need at least C# 11 or 12 for AppServices to work.

Issue #174, opened 2022-06-28
Update version of Uno being used.


mrlacey:
Labs projects are currently targetting Uno 4.2, but 4.3 is out (and 4.4 is due soon)

Uno.UI (v4.2.6) includes a bug that is blocking  #97 
However, updating to v 4.3.8 (the latest stable version) causes `XamlNamedPropertyRelayGenerator` to fail / fall silently--presumably due to the change in the way that Uno uses SourceGenerators.

Also worth confirming/investigating is that as Uno now uses the same sort of Source Generators (SGs) as Labs, does this have any issues for deterministic order? i.e. The Labs SGs rely on the output of the Uno SGs. Can they always be guaranteed to run in the desired order?

Adding a strategy & process for how and when to update to newly released Uno versions should also be planned. Such a plan would naturally extend to all dependencies too.

Arlodotexe:
Looks like only the WPF is having issues when upgrading Uno.

I've tried to update us to the latest version of the Uno templates, but ran into this issue
- https://github.com/unoplatform/uno/issues/9380

Arlodotexe:
Closed in #258

PR #179, opened 2022-06-30
🚧 Parameters for project creation

mrlacey:
For CommunityToolkit/Tooling-Windows-Submodule#65 

What this does:

- A generated experiment will now only include the "classic" version of a control template.
- If the parameter ` --backed ` is included when calling `dotnet new labexp` the generated experiment will include a control that uses a style with a code-behind file, so that it's possible to use `x:Bind`.

This looks to be blocked by a Uno issue and so may need CommunityToolkit/Labs-Windows#174 
Also, for this to be useful, it also needs CommunityToolkit/Labs-Windows#180 to be fixed.


Arlodotexe:
Closing PR as stale. We'll use the code here an example, but the changes are so diverged by now (not even the same repo) that we'll need to redo it.

Issue #212, opened 2022-07-20
🧪 [Experiment] CanvasView - Template and Manipulation enabled Canvas


michael-hawker:
# Approved from Toolkit
See previous work here:

- https://github.com/CommunityToolkit/WindowsCommunityToolkit/discussions/3716
- https://github.com/CommunityToolkit/WindowsCommunityToolkit/issues/854
- Also Related to Experiment #213

## Problem Statement (copied from Discussion):

> I find myself writing this control over and over in various projects and it's never very good. I'd love a control that would be a sibling to ListView or GridView only it was backed by a 2D Canvas instead of an ItemsStackPanel.

## Overview

This experiment adds the following components:

- **CanvasView** a light-weight `ItemsControl` backed by a `Canvas` that allows templating and manipulation of its items (if desired).

## Using

You can try it out via the NuGet Packages here:

- UWP: https://dev.azure.com/dotnet/CommunityToolkit/_artifacts/feed/CommunityToolkit-Labs/NuGet/CommunityToolkit.Labs.Uwp.CanvasView
- WinUI 3: https://dev.azure.com/dotnet/CommunityToolkit/_artifacts/feed/CommunityToolkit-Labs/NuGet/CommunityToolkit.Labs.WinUI.CanvasView

Read more about [Preview Packages here](https://aka.ms/wct/wiki/previewpackages).

CommunityToolkit members can also try it out with Codespaces.

## TO DO
- [ ] Investigate WASM support? (see #174 first)
- [ ] Drift (Delta -> switch to Cumulative) (note blocked by https://github.com/microsoft/microsoft-ui-xaml/issues/6847)
- [ ] Needs to be able to handle some scenarios, like sizing Canvas Width/Height to the contents position/size so it can better support being hosted in a ScrollViewer for a Canvas area that can work larger than a screen. This means keeping track of items which wasn't part of how this initially worked... 🤔

## Implementation Requirements

Not all these items are required to submit a PR. This list is here to help track what is remaining to implement before a technical review and discussion of moving into the main repository can occur.

- [x] Working Prototype
- [ ] Feature Complete
- [ ] Documentation
- [ ] Samples
- [ ] Tests
- [ ] Community Feedback / Usage Testimonies

## Tested Platforms

- [x] UWP
- [x] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

[...]

PR #214, opened 2022-07-20
🧪 Initial import of CanvasView Experiment

michael-hawker:
Long-standing control I've tinkered with on the weekends, finally coming into Labs!

See #212 for details.

Also cleaning up some docs here.

Doesn't seem to be working on WASM, so we need to investigate after #174. Also noticed a drifting when dragging (hadn't done too much with that scenario in my own projects, may have a fix from SizerBase though to try).


michael-hawker:
The merge CI failed here for WinUI 3, I checked and it was an out-of-memory error. So hopefully just transient, the next build passed, so I think we're good. Something to keep an eye on for #174.

Issue #221, opened 2022-07-21
CI Permission issues in PRs from forks 


Arlodotexe:
### Describe the bug

When someone forks the repo and creates a PR, GitHub Actions doesn't run as expected. Instead, it throws errors at us such as:

- [`Resource not accessible by integration`](https://github.com/CommunityToolkit/Labs-Windows/runs/7441466327?check_suite_focus=true#step:12:29) in the `experiment` job.
- [`Error: No test report files were found`](https://github.com/CommunityToolkit/Labs-Windows/runs/7441466211?check_suite_focus=true#step:17:27) in the `Build-WinUI-2` and `Build-WinUI-3` jobs.

### Steps to reproduce

```text
1. Fork the repo
2. Create a new branch
3. Make some changes
4. Open a new PR, merging your forked repo branch back into `main` in the original repo.
5. Wait for CI to fail.
```

### Expected behavior

CI should pass if changes are valid

### Screenshots

![image](https://user-images.githubusercontent.com/9384894/180279373-09c0c868-5904-4bdb-9a55-163f8ec43499.png)

### Code Platform

- [ ] UWP
- [ ] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [ ] Windows 11 21H2 (Build 22000)
- [ ] Other (specify)

### App minimum and target SDK version

- [ ] Windows 10, version 1809 (Build 17763)
- [ ] Windows 10, version 1903 (Build 18362)
- [ ] Windows 10, version 1909 (Build 18363)
- [ ] Windows 10, version 2004 (Build 19041)
- [ ] Other (specify)

### Potentially helpful resources:
- https://github.com/actions/first-interaction/issues/10
-  https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

### Notes to assignees
We should be able to safely fix this by branching main and changing the `build.yml` there, forking the repo, and making a PR to that new branch. That'll cause it to run our modified `build.yml` against the forked branch.

### Help us help you

[...]

Issue #72, opened 2022-04-11
Investigate Testspace.com Module for Test Reporting


michael-hawker:
Looks like open source projects can use Testspace as part of their organization here: https://github.com/marketplace/testspace-com/plan/MDIyOk1hcmtldHBsYWNlTGlzdGluZ1BsYW41OTc0#pricing-and-setup

Once we open the repo we can investigate if we like this any better than the current setup.

PR #223, opened 2022-07-21
Test changes for fixing #221

michael-hawker:


michael-hawker:
Looks like these settings get ignored always from forks. Will close this for now and see if I can test out the TestSpace module from #72.

PR #228, opened 2022-07-25
Niels9001 / Sample app UX improvements

niels9001:

- In TabbedPage mode, all unneccasry chrome is removed (e.g. the Title of the sample)
- Theme toggle button can now be used for samples to verify what your sample looks like in light/dark mode
- Responsiveness bugfixes
- Improving the look and feel of samples on the control pages to reduce text/clutter.

New TabbedPage:

image

Document page:
NewUX


Arlodotexe (approved):
Looked over, tested locally and in a CodeSpace. Seems okay 👍

PR #246, opened 2022-07-29
[Samples] Minor UX tweaks

niels9001:
This PR updates some default backgrounds/margins in the existing samples and project template.

With #228, we get this by default in the ToolkitSampleRenderer.


Arlodotexe (approved):
lgtm, no obvious issues with these changes.

Issue #256, opened 2022-08-10
Update all projects to use .NET 6 where available


Arlodotexe:
### Describe the bug

The following projects need to be updated to use .NET 6:

- [ ] WPF
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Gtk (linux)
- [ ] WASM

This will require generating a blank Uno project and pulling in the new Heads, adapting each one to be usable within the Labs infrastructure. 

### Steps to reproduce

```text
n/a
```

### Expected behavior

n/a

### Code Platform

- [ ] UWP
- [ ] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [ ] Windows 11 21H2 (Build 22000)
- [ ] Other (specify)

### App minimum and target SDK version

- [ ] Windows 10, version 1809 (Build 17763)
- [ ] Windows 10, version 1903 (Build 18362)
- [ ] Windows 10, version 1909 (Build 18363)
- [ ] Windows 10, version 2004 (Build 19041)
- [ ] Other (specify)

### Help us help you

Yes, but only if others can assist.

Arlodotexe:
The WPF head was removed from the "all" solution in #258 due to various issues with Uno/WPF/MSBuild. Tickets have been filed, these can be added back once these are resolved.

Arlodotexe:
Visual Studio 17.3 just came out of preview, and comes preinstalled with the .NET 6 MAUI workloads required by Uno.

https://devblogs.microsoft.com/visualstudio/visual-studio-2022-17-3-is-now-available/

Arlodotexe:
XAML Styler should also be updated to latest version, which includes support for .NET 6: https://github.com/Xavalon/XamlStyler/pull/379

Issue #257, opened 2022-08-10
Use slngen tooling to generate solution file


Arlodotexe:
### Describe the proposed changes

Right now, we're using a PowerShell + a templated version of our previous (Visual Studio generated) solution file.

This is hard to extend and maintain. Instead, we should investigate using the `slngen` tool instead (see [here](https://microsoft.github.io/slngen/)).

### Additional context

This will make #125 much easier.

michael-hawker:
Looks like the shared projects are going to be an issue with slngen (both for the sample app heads and the tests), from #350:

```
"D:\a\Labs-Windows\Labs-Windows\common\CommunityToolkit.Labs.Shared\CommunityToolkit.Labs.Shared.shproj" (default target) (26) ->
(Build target) -> 
  C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VisualStudio\v17.0\CodeSharing\Microsoft.SharedProject.Common.targets(15,5): error MSB3871: Shared projects cannot be built on their own.  Please either build a project that references this project, or build the entire solution. [D:\a\Labs-Windows\Labs-Windows\common\CommunityToolkit.Labs.Shared\CommunityToolkit.Labs.Shared.shproj]
```

michael-hawker:
Shared projects loading with new version now, though WinAppSDK tests not running, so not sure what's going on there...

michael-hawker:
Huh, @Arlodotexe do running the WinAppSDK tests in the all solution work for you? Wonder if something happened and we need to update to the latest (part of the PR I was doing for #183...).

I tried on another branch without slngen and it didn't seem to work. Will check the CI when I update my PR for #350.

michael-hawker:
Think main limitation now is the platform configuration generation in slngen, see https://github.com/microsoft/slngen/issues/437 and a nice-to-have: https://github.com/microsoft/slngen/issues/438

Our current solution's config for reference in future:

![image](https://user-images.githubusercontent.com/24302614/214959441-d1b485f8-6d2c-4f50-9c81-a534abdcfe1f.png)

![image](https://user-images.githubusercontent.com/24302614/214959472-35afde5c-abd2-49d1-9f9d-4c4325b8464d.png)

ARM/ARM64 missing is the biggest issue.

PR #211, opened 2022-07-20
🧪 Bring over StackedNotificationsBehavior from Toolkit

michael-hawker:
Importing from Toolkit

See Details in #210

Been using as my test/demo for presentation, so been working alright. Want to get initial version in.

Made some additional small tweaks:
- Make links in Markdown clickable
- Remove close buttons from Tabs in experiment mode
- Remove empty data at end of docs (think was causing the rogue 'FrontMatter' to showup)
- tweak template to add blank line needed to render list in documentation
- rev packages to check that working with all updates

PR #258, opened 2022-08-10
Updated Uno packages to latest version, removed WPF head from All solution

Arlodotexe:
This PR updates Uno packages to the latest stable version. 

Due to a mountain of issues caused by the WPF head after updating, @michael-hawker and I have opted to simply remove the head from the "All" solution for now as it's not widely used. 

We can revisit adding a functional WPF head back as part of #256


Arlodotexe:
Removed entire instead of commenting out. #257 should make this very easy to add back later.

michael-hawker:
@Arlodotexe I was updating the testspace config as well, as that's misconfigured. Since we're still having issues with the long path in the other PR, did you want to add that here or do as a separate PR? See this commit: https://github.com/CommunityToolkit/Labs-Windows/pull/211/commits/0ce49c32a0d76c3b74f172b1cddc8045161b5d1a

Arlodotexe:
@michael-hawker I'll just cherry pick that onto a new branch and open a PR quick.

michael-hawker (approved):
Tested out in code spaces LGTM

PR #261, opened 2022-08-12
[Gallery] Adding right-to-left toggle

niels9001:
This PR adds a right-to-left toggle so developers can test how their controls.

rtl

@michael-hawker @Arlodotexe 
FlowDirection does not seem to be supported in Uno - however, when I try to hide the button using:

  
  
  I get an error when trying to run the UWP head ("Visibility not found on AppBarButton"). Any thoughts on why that is?

### DIFF
--- a/common/CommunityToolkit.Labs.Shared/Renderers/ToolkitSampleRenderer.xaml
+++ b/common/CommunityToolkit.Labs.Shared/Renderers/ToolkitSampleRenderer.xaml
@@ -7,6 +7,7 @@
       xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
       xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
       xmlns:renderer="using:CommunityToolkit.Labs.Shared.Renderers"
+      xmlns:wasm="http://uno.ui/wasm"
       xmlns:win="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
       Loaded="ToolkitSampleRenderer_Loaded"
       mc:Ignorable="d">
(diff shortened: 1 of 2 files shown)


Arlodotexe (approved):

Arlodotexe:
@niels9001 Looks like your comment is missing something?

image

niels9001:
> @niels9001 Looks like your comment is missing something?
> 
> image
> 

Ooh, some hidden tags.. fixed now 😄

common/CommunityToolkit.Labs.Shared/Renderers/ToolkitSampleRenderer.xaml
```diff
@@ -113,13 +115,16 @@
                             </AppBarButton.Icon>
                         </AppBarButton>
 
-                        <!--  TO DO: Implement FlowDirection  -->
-                        <!--<AppBarButton Label="Toggle right-to-left"
+                        <!--  wasm:Visibility="Collapsed"  -->
```
michael-hawker, line 118: https://user-images.githubusercontent.com/9384894/184418929-412b13a8-a8ce-4e24-bd90-780742f1e25b.png

@niels9001 that seems weird that Visibility wouldn't be working here (@jeromelaban)? Does `wasm:IsEnabled="False"` work as an alternative?

jeromelaban: Using `wasm:Visibility` would generally require to put `wasm` in the ignorables so that UWP skips the declaration. That said, `wasm` is only available on heads, not libraries (libraries are cross targeted with wasm and skia, so they can't be discriminated), so it depends on where this code is included.

Avid29:
As one of the five developers as that uses their machine in a RTL language I just want to celebrate this being added.

PR #268, opened 2022-09-07
Add a RivePlayer experiment

csmartdalton:
Implements a RivePlayer experiment. RivePlayer is a high level control
for playing a Rive state machine or animation. A state machine's
specific inputs can be declared in XAML as nested properties, and
controlled interactively via code or data binding. Low-level Rive
functionality is provided via the RiveSharp NuGet package.

Rive state machines: https://rive.app/features#state-machine
State machine inputs: https://help.rive.app/editor/state-machine#inputs


labs/RivePlayer/src/RivePlayer.Platform.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    private void OnPaintSurface(object? sender, SKPaintGLSurfaceEventArgs e)
+    {
+        this.PaintNextAnimationFrame(e.Surface, e.BackendRenderTarget.Width, e.BackendRenderTarget.Height);
+    }
+#endif
+
+#if HAS_UNO
+    // FrameworkElement has a method named Dispose in Uno, which prevents us from implementing
+    // IDisposable.Dispose. We just have to wait until the destructor runs.
```
csmartdalton, line 39: Any C# experts know how to properly resolve this? I've tried:

```
void IDisposable.Dispose()
  > Rename 'RivePlayer.System.IDisposable.Dispose' to 'Dispose' and ensure that it is declared as public and sealed

public sealed void Dispose()
  > 'RivePlayer.Dispose()' hides inherited member 'FrameworkElement.Dispose()'. Use the new keyword if hiding was intended.

public sealed new void Dispose()
  > 'RivePlayer.Dispose()' cannot be sealed because it is not an override 
  > The member 'RivePlayer.Dispose()' does not hide an accessible member. The new keyword is not required.
  > The member 'RivePlayer.Dispose()' does not hide an accessible member. The new keyword is not required.

```

This works:

```
public new void Dispose()
```

But I suspect with the `new` keywork it actually isn't overriding `IDisposable.Dispose`.

Maybe Uno's `FrameworkElement` sends out a notification we could register to?

We get a warning if RivePlayer doesn't inherit from IDisposable because it owns disposable members.

michael-hawker: @jeromelaban is this something you can help clarify for us?

labs/RivePlayer/src/CommunityToolkit.Labs.WinUI.Rive.RivePlayer.csproj
```diff
@@ -0,0 +1,41 @@
...
+    <PackageReference Include="SkiaSharp.Views" Version="2.88.1" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(IsWinAppSdk)' == 'true'">
+    <PackageReference Include="SkiaSharp.Views.WinUI" Version="2.88.1" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(IsUno)' == 'true'">
+    <PackageReference Include="SkiaSharp.Views.Uno" Version="2.88.2" />
+  </ItemGroup>
```
michael-hawker, line 38: These should be in the `Dependencies.props` file vs. here. There's 4 specific packages that should be needed. (There'll be one for Uno + WindowsAppSDK that's missing here for instance.)

labs/RivePlayer/src/RivePlayer.cs
```diff
@@ -0,0 +1,287 @@
...
+using RiveSharp;
+using SkiaSharp;
+using System.Collections.Concurrent;
+using System.Net.Http;
+using Windows.Storage.Streams;
+using Windows.Storage;
+
+namespace CommunityToolkit.Labs.WinUI.Rive;
+
+// This is a high level XAML control for playing a Rive state machine or animation. The state
```
michael-hawker, line 14: Should look at [xmldoc](https://docs.microsoft.com/dotnet/csharp/language-reference/xmldoc/) comments later as they'll better bubble up in VS intellisense.

labs/RivePlayer/src/RivePlayer.Platform.cs
jeromelaban: Types coming from Uno are implementing `Dispose` as a result of .NET iOS and Android bindings, but actually invoking `Dispose` is causing a large number issues with the lifecycle of underlying native instances. 

This is why Uno is explicitly blocking inherited types from overriding `Dispose` as a best effort, and it's also important to note that corresponding WinUI controls do not provide Dispose either, so the code invoking `Dispose` would not work when compiled for Windows (WinUI or UW).

michael-hawker: Thanks @jeromelaban any recommendations for handling this, I know the lifecycle events of loading/unloading are fickle, so trying to unload/re-initialize based on that instead may be tricky.

(Also, we were having issues with `SKSwapChainPanel` in WASM, should that be working there as well? `SKXamlCanvas` seemed fine.)

jeromelaban: Load/Unload is the only way to handle this, unfortunately, in the same way handling native dependencies work with UWP/WinAppSDK. Otherwise you can try relying on the finalizer, but in this case, you may be having to deal memory leaks and thread affinity to the dispatcher.

For SKSwapChainPanel, this is supposed to work properly (an example [here](https://github.com/unoplatform/Uno.Samples/tree/master/UI/SkiaSharpTest)).

michael-hawker: Ah @csmartdalton looks like there's a 2nd nuget package we need for webassembly:

```xml
    <PackageReference Include="SkiaSharp.NativeAssets.WebAssembly" Version="2.88.0-preview.232" />
```

@Arlodotexe do we have a Labs flag we can use for the dependencies for a conditional here to include a package just for Uno WASM?

labs/RivePlayer/src/CommunityToolkit.Labs.WinUI.Rive.RivePlayer.csproj
csmartdalton: Done!

labs/RivePlayer/src/RivePlayer.cs
csmartdalton: Ack, will do!

labs/RivePlayer/src/RivePlayer.Platform.cs
csmartdalton: I tried adding SkiaSharp.NativeAssets.WebAssembly and it looked like it was already bubbling in through one of the SkiaSharp dependencies.

Also, I just got rid of disposable objects for now. We can start adding them back after this has landed and had a chance to start soaking.

labs/RivePlayer/src/Dependencies.props
```diff
@@ -0,0 +1,40 @@
...
+  <!-- WinUI 3 / WinAppSdk -->
+  <ItemGroup Condition="'$(IsWinAppSdk)' == 'true'">
+    <!-- <PackageReference Include="CommunityToolkit.WinUI.UI.Controls.Primitives" Version="7.1.2"/> -->
+    <PackageReference Include="SkiaSharp.Views.WinUI" Version="2.88.1" />
+  </ItemGroup>
+
+  <!-- WinUI 3 / Uno -->
+  <ItemGroup Condition="'$(IsUno)' == 'true' AND '$(WinUIMajorVersion)' == '3'">
+    <!-- <PackageReference Include="Uno.CommunityToolkit.WinUI.UI.Controls.Primitives" Version="7.1.100-dev.15.g12261e2626"/> -->
+    <PackageReference Include="SkiaSharp.Views.Uno" Version="2.88.2" />
```
Arlodotexe, line 38:
```suggestion
    <PackageReference Include="SkiaSharp.Views.Uno.WinUI" Version="2.88.2" />
```

labs/RivePlayer/samples/RivePlayer.Samples/RivePlayer.md
```diff
@@ -0,0 +1,197 @@
...
+In the example above, we nest the `rive:TriggerInput` strictly under the `rive:RivePlayer` markup.
+
+To create a reference to the trigger input named `trigSuccess` in the state machine of this Rive file, we set the `Target` property to that input name. Additionally, the `x:Name` property defines the name for the class so that the button knows what to bind the click event to to "fire" the trigger.
+
+When the button is clicked, it will call into the `SuccessTriggerInput` input class and invoke the `Fire`  method on the `trigSuccess` Rive trigger input where the state machine will respond accordingly.
+
+Additionally in this example, we use the [EventTriggerBehavior](https://github.com/Microsoft/XamlBehaviors/wiki/EventTriggerBehavior) and [CallMethodAction](https://github.com/Microsoft/XamlBehaviors/wiki/CallMethodAction) APIs from the [XAML Behaviors](https://github.com/Microsoft/XamlBehaviors/wiki) package. You can reference this package by adding the following to your `Dependencies.props`:
+
+```xml
+<!-- WinUI 2 / UWP -->
```
michael-hawker, line 162: This and the dependencies.props is an internal thing to Labs, so app developers just need to do one of the packages based on whatever platform they're using.

csmartdalton: Is the updated wording better?

csmartdalton: Is the updated wording better?

labs/RivePlayer/src/RivePlayer.cs
```diff
@@ -0,0 +1,255 @@
...
+                if (file != null && sourceToken == _currentSourceToken)
+                {
+                    IBuffer buffer = await FileIO.ReadBufferAsync(file);
+                    stream = buffer.AsStream();
+                }
+            }
+            if (stream != null && sourceToken == _currentSourceToken)
+            {
+                byte[] data = new byte[stream.Length];
+                stream.Read(data, 0, data.Length);
```
Sergio0694, line 101: This is doing blocking IO on the UI thread, consider using `ReadAsync` passing the token instead.

Sergio0694, line 101: `Read` (and `ReadAsync`) can read less bytes than requested. Consider using a small helper to loop until all bytes have been read and throw if the operation is not successful.

labs/RivePlayer/src/RivePlayer.cs
```diff
@@ -0,0 +1,255 @@
...
+                var file = await StorageFile.GetFileFromApplicationUriAsync(uri);
+                if (file != null && sourceToken == _currentSourceToken)
+                {
+                    IBuffer buffer = await FileIO.ReadBufferAsync(file);
+                    stream = buffer.AsStream();
+                }
+            }
+            if (stream != null && sourceToken == _currentSourceToken)
+            {
+                byte[] data = new byte[stream.Length];
```
Sergio0694, line 100: This line assumes that the length is available, but that might not always be the case, no? 🤔

labs/RivePlayer/src/RivePlayer.cs
```diff
@@ -0,0 +1,255 @@
...
+            if (mat.Invert(out var inverse))
+            {
+                Vec2D artboardPos = inverse * new Vec2D((float)pointerPos.X, (float)pointerPos.Y);
+                handler(artboardPos);
+            }
+        });
+    }
+
+    // _scene is used on the render thread exclusively.
+    Scene _scene = new Scene();
```
Sergio0694, line 169: Nit: properly order members. All fields should be at the top of the file, and readonly fields first.

labs/RivePlayer/src/RivePlayer.cs
```diff
@@ -0,0 +1,255 @@
...
+
+    // Source actions originating from other threads must be funneled through this queue.
+    readonly ConcurrentQueue<Action> sceneActionsQueue = new ConcurrentQueue<Action>();
```
Sergio0694, line 172: Nit: use target-typed `new()` when initializing fields directly.
```suggestion
    readonly ConcurrentQueue<Action> sceneActionsQueue = new();
```

labs/RivePlayer/src/RivePlayer.cs
```diff
@@ -0,0 +1,255 @@
...
+
+    /// <summary>
+    /// This is the render-thread copy of the state machine parameters. They are set via
+    /// <see cref="_sceneActionsQueue"/>, and applied when the rendering thread calls
+    /// <see cref="UpdateScene"/>
+    /// </summary>
+    private string? _artboardName;
+    private string? _stateMachineName;
+
+    private enum SceneUpdates
```
Sergio0694, line 182: Nit: nested types should be declared last inside a type.

labs/RivePlayer/src/RivePlayer.cs
```diff
@@ -0,0 +1,255 @@
...
+    {
+        EnqueueStateMachineInput(() => _scene.SetNumber(name, value));
+    }
+
+    public void FireTrigger(string name)
+    {
+        EnqueueStateMachineInput(() => _scene.FireTrigger(name));
+    }
+
+    private delegate void PointerHandler(Vec2D pos);
```
Sergio0694, line 142: Is a custom delegate really needed here? Wouldn't `Action<Vec2D>` suffice?

Sergio0694 (changes requested):
This is pretty nice! 🙂
Left some random notes while going over the code 😄

labs/RivePlayer/samples/RivePlayer.Samples/RivePlayerCustomSample.xaml.cs
```diff
@@ -0,0 +1,23 @@
...
+/// </summary>
+[ToolkitSample(id: nameof(RivePlayerCustomSample), "Custom control", description: $"A sample for showing how to create and use a {nameof(RivePlayer)} custom control.")]
+public sealed partial class RivePlayerCustomSample : Page
+{
+    public RivePlayerCustomSample()
+    {
+        this.InitializeComponent();
+#if HAS_UNO
+        // The Uno WASM platform seems to have trouble with our "ms-appx://" URI. Give it an http.
+        this.RivePlayer.Source = "https://public.rive.app/community/runtime-files/2244-4463-animated-login-screen.riv";
```
jeromelaban, line 20: This is generally related to the way the `web.config` file is setup (IIS does not serve unknown mime types by default), which prevents `.riv` files to be downloaded properly.

labs/RivePlayer/src/RivePlayer.cs
csmartdalton: TIL again. That's cool!

labs/RivePlayer/samples/RivePlayer.Samples/RivePlayerCustomSample.xaml.cs
csmartdalton: Thanks for explaining that. This makes sense. I added ".riv" files to web.config and the JS console error went away, but the animation still doesn't load. Anything else I'm missing? Navigating to "https://localhost:5001/animated-login-screen.riv" just loads the same XAML page again (doesn't download  .riv file).

michael-hawker:
@Arlodotexe thoughts on how we exclude the other platforms here for now?

```
"D:\a\Labs-Windows\Labs-Windows\Toolkit.Labs.All.sln" (default target) (1:2) ->
"D:\a\Labs-Windows\Labs-Windows\platforms\CommunityToolkit.Labs.Droid\CommunityToolkit.Labs.Droid.csproj" (default target) (6:6) ->
"D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\CommunityToolkit.Labs.WinUI.Rive.RivePlayer.csproj" (default target) (18:83) ->
(CoreCompile target) -> 
  D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\RivePlayer.Platform.cs(9,23): error CS0234: The type or namespace name 'UWP' does not exist in the namespace 'SkiaSharp.Views' (are you missing an assembly reference?) [D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\CommunityToolkit.Labs.WinUI.Rive.RivePlayer.csproj]
  D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\RivePlayer.Platform.cs(87,49): error CS0246: The type or namespace name 'SKPaintGLSurfaceEventArgs' could not be found (are you missing a using directive or an assembly reference?) [D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\CommunityToolkit.Labs.WinUI.Rive.RivePlayer.csproj]


"D:\a\Labs-Windows\Labs-Windows\Toolkit.Labs.All.sln" (default target) (1:2) ->
"D:\a\Labs-Windows\Labs-Windows\platforms\CommunityToolkit.Labs.iOS\CommunityToolkit.Labs.iOS.csproj" (default target) (7:6) ->
"D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\CommunityToolkit.Labs.WinUI.Rive.RivePlayer.csproj" (default target) (18:87) ->
  D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\RivePlayer.Platform.cs(9,23): error CS0234: The type or namespace name 'UWP' does not exist in the namespace 'SkiaSharp.Views' (are you missing an assembly reference?) [D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\CommunityToolkit.Labs.WinUI.Rive.RivePlayer.csproj]
  D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\RivePlayer.Platform.cs(87,49): error CS0246: The type or namespace name 'SKPaintGLSurfaceEventArgs' could not be found (are you missing a using directive or an assembly reference?) [D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\CommunityToolkit.Labs.WinUI.Rive.RivePlayer.csproj]


"D:\a\Labs-Windows\Labs-Windows\Toolkit.Labs.All.sln" (default target) (1:2) ->
"D:\a\Labs-Windows\Labs-Windows\platforms\CommunityToolkit.Labs.macOS\CommunityToolkit.Labs.macOS.csproj" (default target) (8:6) ->
"D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\CommunityToolkit.Labs.WinUI.Rive.RivePlayer.csproj" (default target) (18:91) ->
  D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\RivePlayer.Platform.cs(9,23): error CS0234: The type or namespace name 'UWP' does not exist in the namespace 'SkiaSharp.Views' (are you missing an assembly reference?) [D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\CommunityToolkit.Labs.WinUI.Rive.RivePlayer.csproj]
  D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\RivePlayer.Platform.cs(87,49): error CS0246: The type or namespace name 'SKPaintGLSurfaceEventArgs' could not be found (are you missing a using directive or an assembly reference?) [D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\CommunityToolkit.Labs.WinUI.Rive.RivePlayer.csproj]


"D:\a\Labs-Windows\Labs-Windows\Toolkit.Labs.All.sln" (default target) (1:2) ->
"D:\a\Labs-Windows\Labs-Windows\platforms\CommunityToolkit.Labs.Skia.Gtk\CommunityToolkit.Labs.Skia.Gtk.csproj" (default target) (9:6) ->
"D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\CommunityToolkit.Labs.WinUI.Rive.RivePlayer.csproj" (default target) (18:95) ->
  D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\RivePlayer.Platform.cs(9,23): error CS0234: The type or namespace name 'UWP' does not exist in the namespace 'SkiaSharp.Views' (are you missing an assembly reference?) [D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\CommunityToolkit.Labs.WinUI.Rive.RivePlayer.csproj]
  D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\RivePlayer.Platform.cs(82,49): error CS0246: The type or namespace name 'SKPaintSurfaceEventArgs' could not be found (are you missing a using directive or an assembly reference?) [D:\a\Labs-Windows\Labs-Windows\labs\RivePlayer\src\CommunityToolkit.Labs.WinUI.Rive.RivePlayer.csproj]
```

I think they have different namespaces in the SkiaSharp package for each platform, so even if we don't want to test them yet, we could see if they'll compile by at least having conditional compiles to swap out the right references, eh?

labs/RivePlayer/samples/RivePlayer.Samples/RivePlayerCustomSample.xaml.cs
jeromelaban: This still means that the extension is filtered somehow `.riv`. Changing the `web.config` should have been enough. Does it work if you call it `.dat` ?

jeromelaban:
> I think they have different namespaces in the SkiaSharp package for each platform, so even if we don't want to test them yet, we could see if they'll compile by at least having conditional compiles to swap out the right references, eh?

That's correct, the namespace has changed with WinUI 3, but I don't think you'll have a way to change this with WinUI unfortunately.

michael-hawker:
> > I think they have different namespaces in the SkiaSharp package for each platform, so even if we don't want to test them yet, we could see if they'll compile by at least having conditional compiles to swap out the right references, eh?
> 
> That's correct, the namespace has changed with WinUI 3, but I don't think you'll have a way to exclude it in-XAML with WinUI unfortunately.

These are created in the `.cs` files, so should be doable, think there's just a disconnect here somewhere around what's needed where for each platform, I know a lot of this stuff isn't documented that clearly across the ecosystem for the Windows App SDK.

That's also why we're aligning in 8.0 to the same namespace across UWP and WinUI for the Toolkit to reduce all these headaches when migrating platforms and sharing code/samples. Something we should suggest maybe for SkiaSharp in the future too? :) @mattleibow

labs/RivePlayer/samples/RivePlayer.Samples/RivePlayerCustomSample.xaml.cs
csmartdalton: Calling it `.dat` does not work and doesn't print an error to the web developer console either.

Should I be able to navigate to this file somewhere when the local server is running? "https://localhost:5001/animated-login-screen.dat" maybe?

.github/workflows/build.yml
```diff
@@ -267,6 +267,9 @@ jobs:
         shell: pwsh
         run: ./GenerateAllSolution.ps1
 
+      - name: Install ninja
+        run: apt-get install ninja-build
```
michael-hawker, line 271:
```suggestion
      - name: Install ninja for WASM native dependencies
        run: sudo apt-get install ninja-build
```

Add sudo command for install (and improve description). @Arlodotexe will this effect codespaces?

labs/RivePlayer/src/RivePlayer.Platform.cs
```diff
@@ -0,0 +1,106 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if WINDOWS_UWP || (WINUI2 && HAS_UNO_WASM)
```
michael-hawker, line 5:
```suggestion
#if WINDOWS_UWP || (WINUI2 && HAS_UNO)
```

Think this should just be `HAS_UNO`?

Arlodotexe: You should be able to shorten this to just:
```suggestion
#if !WINAPPSDK
```

csmartdalton: `!WINAPPSDK` seemed to build locally for WinUI2 and WinUI3 both. Let's see if it works here!

labs/RivePlayer/src/RivePlayer.xaml
```diff
@@ -0,0 +1,28 @@
...
+    <Style TargetType="Rive:RivePlayer">
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="Rive:RivePlayer">
+                    <ContentPresenter x:Name="SkiaSurface"
+                                      Padding="{TemplateBinding Padding}"
+                                      HorizontalContentAlignment="Stretch"
+                                      VerticalContentAlignment="Stretch"
+                                      win:AutomationProperties.AccessibilityView="Raw"
+                                      Background="{TemplateBinding Background}"
```
niels9001, line 18: Nit: should we provide a default BackgroundBrush? I noticed that at times it took a moment to load the content, so having some sort of indication that something is loading might be helpful? (A ProgressRing could be an option as well)?

labs/RivePlayer/samples/RivePlayer.Samples/RivePlayerCustomSample.xaml
```diff
@@ -0,0 +1,67 @@
...
+            <rive:NumberInput Target="numLook"
+                              Value="{x:Bind NumLook.Value, Mode=OneWay}" />
+            <rive:BoolInput Target="isHandsUp"
+                            Value="{x:Bind IsHandsUp.IsChecked, Mode=OneWay}" />
+            <rive:TriggerInput x:Name="FailureTrigger"
+                               Target="trigFail" />
+            <rive:TriggerInput x:Name="SuccessTrigger"
+                               Target="trigSuccess" />
+        </rive:RivePlayer>
+        <StackPanel Grid.Column="1"
```
niels9001, line 35: We should move these to the SampleOptions. but I don't think we support all of these types of options yet. So something to fix later on :).

niels9001:
Awesome work :).. so exciting to see this landing!

@csmartdalton does Rive have some sort of demo gallery where users could try out different animations? Would be nice to mention that in the documentation maybe so they could play around with those locally?

michael-hawker (approved):
Tested that we can still build other experiments individually in Codespaces, so we'll fix that forward with WASM improvements to Uno once .NET 7 ships later next month when we work on #283.

.github/workflows/build.yml
Arlodotexe: `./.github/workflows/build.yml` is only used by GitHub Actions. This should be safe to do in Actions, according to the [docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners).

> The Linux and macOS virtual machines both run using passwordless sudo. When you need to execute commands or install tools that require more privileges than the current user, you can use sudo without needing to provide a password.

To install this in Codespaces, modify the Docker config at `./.devcontainer/Dockerfile`.

PR #271, opened 2022-09-08
feat: Update Uno Fluent Icons

MartinZikmund:
Updates the icon font in sample apps to the latest Uno Fluent Icons (based on Windows 11 iconography style)

Before vs. after
image


niels9001:
@MartinZikmund Nice work 😊! Added a before vs. after image to your post.

image.

Somehow it seems that the icons in the NavigationView seem to be cut-off a bit.. might be an Uno.NavigationView issue (like a hardcoded height or margin/padding that was set with MDL2 Icons in mind)?

MartinZikmund:
@niels9001 Indeed that looks like an Uno `NavigationView` issue, you are seeing the cut-off on WASM? I will look into it

MartinZikmund:
@niels9001 Hi, can you please verify the last force push fixed the issue for you?

PR #275, opened 2022-09-15
 [Feature] Marquee Text Control

Avid29:
Adds a new MarqueeText control that scrolls text across the screen as a Marquee.

https://user-images.githubusercontent.com/9384045/190292642-e187f29e-8024-4906-bcb0-368544dafc85.mp4


Avid29:
Sorry for the resubmit. I realized the base branch name was wrong

labs/MarqueeText/src/MarqueeText.xaml
```diff
@@ -0,0 +1,103 @@
...
+                                           FontStyle="{TemplateBinding FontStyle}"
+                                           FontWeight="{TemplateBinding FontWeight}"
+                                           Foreground="{TemplateBinding Foreground}"
+                                           Text="{TemplateBinding Text}"
+                                           TextDecorations="{TemplateBinding TextDecorations}" />
+                                <TextBlock x:Name="Segment2"
+                                           CharacterSpacing="{TemplateBinding CharacterSpacing}"
+                                           FontFamily="{TemplateBinding FontFamily}"
+                                           FontSize="{TemplateBinding FontSize}"
+                                           FontStretch="{TemplateBinding FontStretch}"
```
niels9001, line 90:
```suggestion
                                           win:FontStretch="{TemplateBinding FontStretch}"
```

labs/MarqueeText/src/MarqueeText.xaml
```diff
@@ -0,0 +1,103 @@
...
+                            <StackPanel x:Name="MarqueePanel"
+                                        Orientation="Horizontal">
+                                <StackPanel.RenderTransform>
+                                    <TranslateTransform x:Name="MarqueeTransform" />
+                                </StackPanel.RenderTransform>
+                                <TextBlock x:Name="Segment1"
+                                           CharacterSpacing="{TemplateBinding CharacterSpacing}"
+                                           FontFamily="{TemplateBinding FontFamily}"
+                                           FontSize="{TemplateBinding FontSize}"
+                                           FontStretch="{TemplateBinding FontStretch}"
```
niels9001, line 80:
```suggestion
                                           win:FontStretch="{TemplateBinding FontStretch}"
```

labs/MarqueeText/src/MarqueeText.xaml
```diff
@@ -0,0 +1,103 @@
+﻿<!--  Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information.  -->
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+                    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
```
niels9001, line 3:
```suggestion
                    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
                    xmlns:win="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
```

niels9001:
@Avid29 Awesome work :)! FYI, I've added the win: prefix for FontStretch (as it's not supported for Uno) so that the build will hopefully pass!

Avid29:
So the current issue appears to be that in Uno (and only in Uno) ``Timeline`` is ``IDisposible``, and therefore the ``MarqueeText`` control must also be ``IDisposible`` since it contains a ``Storyboard``. However in UWP there's no reason to make the control ``IDisposible``, so I'm really not tempted to just add that interface for no reason. For now, I have conditionally attached a ``IDisposible`` for Uno, but I'm not too happy with this solution.
image


Any better ideas?

michael-hawker:
@Avid29 was there a warning message about this or something somewhere calling this out?

@jeromelaban is this `IDisposable` message a red herring, are we missing something else instead?

jeromelaban:
Some objects are marked as disposable for internal constraints, but it's not something we can change right away as it may be a breaking change (we'll take note to add it as such for our next major). Having to add `IDisposable` here is likely a result of some analyzer that can be disabled for Uno targets specifically.

Avid29:
Hmm. That's frustrating. It fails with ``IDisposable`` because it's redundant, and it fails without ``IDisposible`` because it's missing...

Arlodotexe:
> Having to add IDisposable here is likely a result of some analyzer that can be disabled for Uno targets specifically.

@Avid29 See if you can find which analyzer is doing that. Since `IDisposable` is implemented as an internal constraint to Uno, we can effectively ignore this and disable the analyzer on uno platforms.

Arlodotexe:
> > Having to add IDisposable here is likely a result of some analyzer that can be disabled for Uno targets specifically.
> 
> @Avid29 See if you can find which analyzer is doing that. Since `IDisposable` is implemented as an internal constraint to Uno, we can effectively ignore this and disable the analyzer on uno platforms.

Partially fixed in #377, but since MarqueeText still owns the field, we need to manually suppress the error in this file. There's no way to suppress this problem globally for all files.

Arlodotexe (approved):

Avid29:
@niels9001 I believe this is just waiting on your approval, right @Arlodotexe?

components/MarqueeText/samples/MarqueeText.md
```diff
@@ -0,0 +1,24 @@
...
+subcategory: StatusAndInfo
+discussion-id: 231 
+issue-id: 0
+---
+
+# MarqueeText
+
+For more information about this experiment see:
+
+- Discussion: https://github.com/CommunityToolkit/Labs-Windows/discussions/231
```
niels9001, line 18: This can be removed, we have the discussion button that will take users there (using the discussion ID).

components/MarqueeText/samples/MarqueeText.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Issue: TODO: PASTE LINK HERE
+
+The MarqueeText control allows text to scroll in a marquee fashion. The control is heavily templated and many changes can be made by modifying the style.
```
niels9001, line 22: Maybe mention the looping/speed properties?

niels9001:
> @niels9001 I believe this is just waiting on your approval, right @Arlodotexe?

2 minor comments.. otherwise looking great 😁! Awesome control!

components/MarqueeText/samples/MarqueeText.md
niels9001, line 18:
```suggestion
```

components/MarqueeText/samples/MarqueeText.md
```diff
@@ -0,0 +1,24 @@
...
+issue-id: 0
+---
+
+# MarqueeText
+
+For more information about this experiment see:
+
+- Discussion: https://github.com/CommunityToolkit/Labs-Windows/discussions/231
+
+- Issue: TODO: PASTE LINK HERE
```
niels9001, line 20:
```suggestion
```

components/MarqueeText/samples/MarqueeText.md
niels9001, line 22: Maybe mention the Behavior / Direction / Speed / RepeatBehavior here?

Avid29: Yeah, I kind of ignored the markdown file. I just pushed a new commit with some more full docs, but I couldn't get multiple demo pages without the compiler blowing up 🤷‍♂️

niels9001: Changes looking great :)! Thanks!

niels9001 (approved):

Avid29:
In case anyone finds their way here, the `MarqueeText` control has been replaced by simply the `Marquee` control as of this PR:
https://github.com/CommunityToolkit/Labs-Windows/pull/732

PR #281, opened 2022-10-06
Applied workaround for failing Xamarin.Mac builds

Arlodotexe:
MacOS builds are failing on the `main` branch:
https://github.com/CommunityToolkit/Labs-Windows/actions/runs/3198993878

This PR fixes this by applying this workaround:
https://developercommunity.visualstudio.com/t/XamarinMac-binaries-are-missing-in-173/10164443#T-N10164676

### DIFF
--- a/common/Labs.MultiTarget.props
+++ b/common/Labs.MultiTarget.props
@@ -53,4 +53,14 @@
     <EmbeddedResource Remove="obj\bin\**" />
     <UpToDateCheckInput Remove="obj\bin\**" />
   </ItemGroup>
+
+  <!-- Workaround for Mac: https://developercommunity.visualstudio.com/t/XamarinMac-binaries-are-missing-in-173/10164443#T-N10164676 10/6/22 -->
+  <ItemGroup Condition=" $(TargetFramework.StartsWith('Xamarin.Mac')) ">
+    <Reference Include="Xamarin.Mac">
+      <HintPath Condition="Exists('c:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\IDE\Extensions\Xamarin.VisualStudio')">c:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\IDE\Extensions\Xamarin.VisualStudio\Xamarin.Mac.dll</HintPath>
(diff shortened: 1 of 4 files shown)


Arlodotexe:
Looks like the supplied workaround hasn't fixed the issue. @jeromelaban any ideas? It's being applied to all multitargeted library projects.

common/Labs.MultiTarget.props
```diff
@@ -53,4 +53,14 @@
     <EmbeddedResource Remove="obj\bin\**" />
     <UpToDateCheckInput Remove="obj\bin\**" />
   </ItemGroup>
+
+  <!-- Workaround for Mac: https://developercommunity.visualstudio.com/t/XamarinMac-binaries-are-missing-in-173/10164443#T-N10164676 10/6/22 -->
+  <ItemGroup Condition=" $(TargetFramework.StartsWith('Xamarin.Mac')) ">
```
jeromelaban, line 58: See here:
https://github.com/CommunityToolkit/Labs-Windows/blob/ed6bd23523770f34ebf6b41304dd257374fe1f04/common/Labs.TargetFrameworks.All.props#L11

```suggestion
  <ItemGroup Condition=" $(TargetFramework.StartsWith('xamarinmac')) ">
```

jeromelaban:
Interesting. The issue is still there. I wonder if the fix is different for github actions agents. It would be a good idea to publish the binlogs of the build for troubleshooting.

Arlodotexe:
@jeromelaban Binlogs from GH Actions are available [here](https://github.com/CommunityToolkit/Labs-Windows/suites/8673877158/artifacts/390687343)

jeromelaban:
I've pushed a change specific for the macOS head, it's a "legacy" project which does not pick up the Directory.Build.props above.

michael-hawker:
Thanks @jeromelaban for the help here! @Arlodotexe did we want to keep the binlog settings (that's going to slow it down a bit, eh? and take up more space?)

Arlodotexe:
> Thanks @jeromelaban for the help here! @Arlodotexe did we want to keep the binlog settings (that's going to slow it down a bit, eh? and take up more space?)

Let's remove it if it's not needed.

michael-hawker (approved):

jeromelaban:
Binlogs are generally quite useful for build troubleshooting. We (In Uno) tend to reduce most textual build logging and keep the binlogs which are far more detailed and more performant. You may want to comment out the task that does the artifacts copy, just in case :)

PR #285, opened 2022-10-16
[SettingsControls] UI fixes

niels9001:
This PR fixes the following two bugs:

- If the HorizontalContentAlignment was explicitly set on the Content, this would not be applied.
- ToggleSwitch labels wouldn't correctly align to the right.

Togglefix


michael-hawker (approved):
Cool, I'll test the new package out in my app test. 🙂

We should probably think about having example cards with the different types of controls in the samples as well as we expand those.

After I get an initial preview build for the 7.1.3 hotfix, I'll try to add a sample for using `ItemsSource` and try and figure out what was going on there.

Issue #216, opened 2022-07-20
🧪 [Experiment] SettingsCard & SettingsExpander


niels9001:
# Approved from Toolkit

- https://github.com/CommunityToolkit/Labs-Windows/discussions/129

## Problem Statement:

THere are currently no controls in WinUI or the Toolkit that allows for creating consistent settings experiences that can be found across Windows 11. This experiment introduces a SettingsCard component that allows to display simple cards.
SettingsExpander will be an follow-up experiment to introduce collapsible cards.

## Overview

This experiment adds the following components:

- **SettingsCard** a simple card component that allows for displaying a setting. The `IsClickEnabled` property can be used to turn it into a Button-like control.
- **SettingsExpander** a control that uses the same properties as `SettingsCard`, and `SettingsCard` can be used to set the SettingsExpander.Items. Binding is also supported.
![SettingsExperiment](https://user-images.githubusercontent.com/9866362/179772289-abc9bb1a-bb06-49df-8fe4-1628e4dd3852.gif)

https://user-images.githubusercontent.com/9866362/223989590-611a26f6-74ce-4c6e-9bb6-055df743a5c5.mp4

## Using

You can try it out via the NuGet Packages here:

- UWP: https://dev.azure.com/dotnet/CommunityToolkit/_artifacts/feed/CommunityToolkit-Labs/NuGet/CommunityToolkit.Labs.Uwp.SettingsControls
- WinUI 3: https://dev.azure.com/dotnet/CommunityToolkit/_artifacts/feed/CommunityToolkit-Labs/NuGet/CommunityToolkit.Labs.WinUI.SettingsControls

[SettingsCard (doc + samples)](https://github.com/CommunityToolkit/Labs-Windows/blob/main/components/SettingsControls/samples/SettingsCard.md)
[SettingsExpander (doc + samples)](https://github.com/CommunityToolkit/Labs-Windows/blob/main/components/SettingsControls/samples/SettingsExpander.md)

Read more about [Preview Packages here](https://aka.ms/wct/wiki/previewpackages).

CommunityToolkit members can also try it out with Codespaces.

## TO DO
- [ ] Tests

## Implementation Requirements

Not all these items are required to submit a PR. This list is here to help track what is remaining to implement before a technical review and discussion of moving into the main repository can occur.

- [x] Working Prototype
- [x] Feature Complete
- [x] Documentation
- [x] Samples
- [ ] Tests
- [ ] Community Feedback / Usage Testimonies

[...]

PR #290, opened 2022-10-21
Fix SettingsExpander ItemsSource scenario

michael-hawker:
# Tied to Experiment #129 #216 

Switch SettingsExpander to use ItemsRepeater and inherit from Control instead of ItemsControl

This is the WinUI approach TabView/NavigationView uses and allows us to specify the entire Container in the DataTemplate:

```xml
        <labs:SettingsExpander Description="The SettingsExpander can use ItemsSource to define its Items."
                               Header="Settings Expander with ItemsSource"
                               ItemsSource="{x:Bind MyDataSet}">
            <labs:SettingsExpander.ItemTemplate>
                <DataTemplate x:DataType="local:MyDataModel">
                    <labs:SettingsCard Description="{x:Bind Info}"
                                       Header="{x:Bind Name}">
                        <HyperlinkButton Content="{x:Bind LinkDescription}"
                                         NavigateUri="{x:Bind Url}" />
                    </labs:SettingsCard>
                </DataTemplate>
            </labs:SettingsExpander.ItemTemplate>
        </labs:SettingsExpander>
```

This is needed for the ItemSource scenario of SettingsExpander to work properly. Otherwise ItemsControl only sees the data item and sticks the container in another container, like this:

image

With this change the desired behavior works properly now:

image

This PR also split out the samples for more delineation in the future and room for a full complete settings page sample.

## Notes:

- [ ] We could also inherit from ContentControl instead of Control like NavigationView, but this was a simpler first step.

- [ ] Some ItemsControl properties/methods are still missing, but initial approach seems to work and resolve issues with our same desired API surface.

- [ ] New issue: Expander is shifting position when expanding for some reason now? (don't think it did that before)


labs/SettingsControls/src/SettingsExpander/SettingsExpander.ItemsControl.cs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    private void ItemsRepeater_ElementPrepared(MUXC.ItemsRepeater sender, MUXC.ItemsRepeaterElementPreparedEventArgs args)
+    {
+        if (ItemContainerStyleSelector != null &&
+            args.Element is FrameworkElement element &&
+            element.ReadLocalValue(FrameworkElement.StyleProperty) == DependencyProperty.UnsetValue)
+        {
+            // TODO: Get item from args.Index?
+            element.Style = ItemContainerStyleSelector.SelectStyle(null, element);
```
michael-hawker, line 69: Wonder if DataContext is set already here, should check, as that'd be just easier?

michael-hawker:
Random build failure again:

```
"D:\a\Labs-Windows\Labs-Windows\Toolkit.Labs.All.sln" (default target) (1:2) ->
"D:\a\Labs-Windows\Labs-Windows\platforms\CommunityToolkit.Labs.Uwp\CommunityToolkit.Labs.Uwp.csproj" (default target) (10:6) ->
(BuildNativePackage target) -> 
  C:\Users\runneradmin\.nuget\packages\microsoft.net.native.compiler\2.2.10-rel-29722-00\tools\Microsoft.NetNative.targets(805,5): error : Error: NUTC300F:Internal Compiler Error: Native compilation failed due to out of memory error [D:\a\Labs-Windows\Labs-Windows\platforms\CommunityToolkit.Labs.Uwp\CommunityToolkit.Labs.Uwp.csproj]
  C:\Users\runneradmin\.nuget\packages\microsoft.net.native.compiler\2.2.10-rel-29722-00\tools\Microsoft.NetNative.targets(805,5): error : ILT0005: 'C:\Users\runneradmin\.nuget\packages\runtime.win10-x64.microsoft.net.native.compiler\2.2.10-rel-29722-00\tools\x64\ilc\Tools\nutc_driver.exe @"D:\a\Labs-Windows\Labs-Windows\platforms\CommunityToolkit.Labs.Uwp\obj\x64\Release\ilc\intermediate\MDIL\CommunityToolkit.Labs.Uwp.rsp"' returned exit code 1 [D:\a\Labs-Windows\Labs-Windows\platforms\CommunityToolkit.Labs.Uwp\CommunityToolkit.Labs.Uwp.csproj]
```

Some of this was going to be resolved by the work I was trying to do on warnings for the long path of #211 as well.

niels9001:
@michael-hawker So the jumping issue was gone when I removed the ScrollViewer in the Expander samples. The ToolkitSampleRenderer already has a ScrollViewer for these situations:

SettingsExpanderScrollBug

But it's not there when using a ScrollViewer on main - so there might be some issue still. Will check if I can find that :)

Overall, I think we need to re-XAML the sample renderer layout to avoid this quirks and think about how we want to deal with an overrideable MaxHeight and all that!

Sergio0694:
@michael-hawker The .NET Native toolchain is a transitive dependency from [Microsoft.NETCore.UniversalWindowsPlatform](https://www.nuget.org/packages/Microsoft.NETCore.UniversalWindowsPlatform/). You'll want to update to the latest and also ensure you're using the 64 bit binder. For that, set `<OutOfProcPDB>true</OutOfProcPDB>` in your .csproj file.

michael-hawker:
Using ItemsRepeater blocked by this WinUI issue: https://github.com/microsoft/microsoft-ui-xaml/issues/3842

niels9001 (approved):
Tested, jumping is gone :). Great work!

PR #305, opened 2022-11-09
Fixed text selection for code display on WASM

Arlodotexe:
This PR contains a quick fix which enables text selection in WASM for the `TextBlock`s which render code.


michael-hawker (approved):
Tested in Codespaces, works great! (Should enable in the future for the documentation sections too, but that may be trickier with the new approach in #308, will investigate later.)

PR #307, opened 2022-11-09
Update RivePlayer to refresh with DispatcherQueueTimer

csmartdalton:
DispatcherQueueTimer is cross-compatible and cleaner/more correct than
the custom timer from before.


labs/RivePlayer/src/AnimationTimer.cs
```diff
@@ -52,7 +52,8 @@ public void Stop()
 
     private async void InvalLoopAsync(int continuationToken)
     {
-        while (continuationToken == _invalLoopContinuationToken)
+        while (continuationToken == _invalLoopContinuationToken &&
+               rivePlayerRef.TryGetTarget(out var rivePlayer))
```
michael-hawker, line 56: This is still going to just check for the reference once and then hold it in the local loop which will keep the control loaded. @Sergio0694 suggestions on how to break the handle cycle here with the render loop?

michael-hawker: Ah, was just thinking we could split this into two parts (the loop and the action):

```cs
    private async void InvalLoopAsync(int continuationToken)
    {
        while (continuationToken == _invalLoopContinuationToken)
        {
             PerformLoopAction();
             await Task.Delay(timeSpanPerFrame);
        }
    }
    
    private void PerformLoopAction()
    {
        if (rivePlayerRef.TryGetTarget(out var rivePlayer))
        {
            if (!rivePlayer.IsLoaded)
            {
                Stop():
                return;
            }
            
            rivePlayer.InvalidateAnimation();
        }
        else
        {
            Stop();
        }
    }
```

labs/RivePlayer/src/AnimationTimer.cs
```diff
@@ -54,12 +54,29 @@ private async void InvalLoopAsync(int continuationToken)
     {
         while (continuationToken == _invalLoopContinuationToken)
```
michael-hawker, line 55: @csmartdalton - @Sergio0694 suggested we use a `DispatcherQueueTimer` here for better performance, you can see a similar scenario we have in a Toolkit control here: https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/2233ac655ff24bc77a9fa23f5f70b0434230f342/Microsoft.Toolkit.Uwp.UI.Controls.Core/RotatorTile/RotatorTile.cs#L389-L394

Basically the `Tick` event handler becomes our new method below and you can get rid of most of the other code here in your class. 🙂 Let us know if you need a hand.

csmartdalton: I tried using `DispatcherQueueTimer` but it looks like it's not available everywhere:

notavailable

Is there a way to use it in net6.0 or would it only be a partial solution?

michael-hawker: @csmartdalton it's available, it's just a different namespace/mapping which should be handled by Labs, so you should be able to ignore the VS warning (I'm not sure why it doesn't understand that probably as it's looking for the fully qualified type name or something).

Filed an issue with the docs here: https://github.com/MicrosoftDocs/winrt-api/issues/2309

michael-hawker: Filed https://github.com/CommunityToolkit/Labs-Windows/issues/331 for tracking.

csmartdalton: Thanks for the offline help @michael-hawker ! It's working now.

Arlodotexe (approved):

PR #308, opened 2022-11-09
Sample App Issue/Discussion Links and Markdown Rendering

### LINKED ISSUE #151: [Sample App] Markdown Renderer Polyfill
_Split out from #93, Follow-on from PR #85_

## Problem Statement

Our current documentation is shown in raw Markdown instead of nice documentation.

We can use the Toolkit's `MarkdownTextblock` to resolve this, _except on WASM_. So we need a wrapper to make this work everywhere (in the sample app, not something we want to ship as a control).

## Technical Background

In #85, I wrote code to split out the markdown snippets and inject the displayed samples in the `ToolkitDocumentationRenderer`:

https://github.com/CommunityToolkit/Labs-Windows/blob/8f2380383bd82d3620ad65e4e085b15dab92a855/common/CommunityToolkit.Labs.Shared/Renderers/ToolkitDocumentationRenderer.xaml.cs#L121-L141

As well as a Template Selector to pick the right template to use:

https://github.com/CommunityToolkit/Labs-Windows/blob/8f2380383bd82d3620ad65e4e085b15dab92a855/common/CommunityToolkit.Labs.Shared/Renderers/ToolkitDocumentationRenderer.xaml#L13-L29

So, all we should have to do is update the `DocumentTemplate` to leverage this new control that understands how to display the Markdown snippet.

## Solution

Fortunately, @nickrandolph started a markdown based solution using [marked.js](https://marked.js.org/) for WASM that we can use for the individual blurbs from the MVVM Toolkit Sample App, see here: https://github.com/CommunityToolkit/MVVM-Samples/pull/38/commits/911b3551d6034dccd721af57ebbd502a4130b5aa~
  
And he's given us permission to leverage that code from that PR which didn't get merged:

image

I'm not sure if his version used WebView/marked even on native platform, so that may be a difference, but the general approach and code here should be handy to reference.

We'll probably want to update the Marked version referenced from the original PR there (and not sure if they included minified version).

We'll also need to a [Third Party Notice file like we have in the toolkit](https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/main/ThirdPartyNotices.txt) since we'll be including the source directly as it's JavaScript.

## Open Questions

- [ ] Do we worry about supporting DocFX or the new [GitHub flavored](https://github.com/orgs/github-community/discussions/16925#discussioncomment-2904076) Note/Warning sections at the moment? We should eventually, but we did this with a specific [custom renderer in the toolkit](https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/main/Microsoft.Toolkit.Uwp.SampleApp/Controls/SampleAppMarkdownRenderer.cs) and not sure of marked.js support (though they also [have custom rendering](https://marked.js.org/using_pro#renderer)).
  - [ ] Was thinking would be useful for 'preview/experimental' banner at top of experiments, but maybe we inject that on top of everything manually or in the side-bar of the app? @niels9001 thoughts?

michael-hawker:
Adds:
- Replaces buttons at top of samples with 'Discussion' and 'Tracking Issue' which link to the discussion/issue for the experiment here on GitHub. Uses new metadata fields on top of frontmatter for docs.
- Markdown rendering via Markdig in WASM via HTML (Fixes #151)

Tested in WASM locally

Known issues for Markdown
- [ ] layout potential mismatch for image loading on Rive page, resizing fixes (need to see if happens without debugger) - this seems random?
- [x] links in markdown files contrast is bad as it's default link blue (forgot to test trying to open them as well)


niels9001:
@michael-hawker I'm still having some WASM issues on my local machine so haven't tested that yet. UWP looked good though!

Removed some more obsolete links on the CanvasLayout and CanvasView page.

common/CommunityToolkit.Labs.Shared/Renderers/Markdown/MarkdownTextBlock.cs
```diff
@@ -16,11 +16,78 @@
...
+            mtb.UpdateText(e.NewValue as string ?? string.Empty);
+        }
+    }
+
+    public MarkdownTextBlock()
+    {
+        Loaded += this.MarkdownTextBlock_Loaded;
+    }
+
+    private void MarkdownTextBlock_Loaded(object sender, RoutedEventArgs e)
```
niels9001, line 54: sender should be nullable?

common/CommunityToolkit.Labs.Shared/Renderers/Markdown/MarkdownTextBlock.cs
```diff
@@ -16,11 +16,79 @@
...
+            mtb.UpdateText(e.NewValue as string ?? string.Empty);
+        }
+    }
+
+    public MarkdownTextBlock()
+    {
+        Loaded += this.MarkdownTextBlock_Loaded;
+    }
+
+    #nullable enable
```
michael-hawker, line 54: @Arlodotexe not sure why I need this here, but VS was giving me a squiggle without it.

common/CommunityToolkit.Labs.Shared/Renderers/Markdown/MarkdownTextBlock.cs
```diff
@@ -16,11 +16,79 @@
     #endif
 #endif
 
+#if __WASM__
+using Markdig;
```
michael-hawker, line 20: @Arlodotexe this seems weird?

image

Why is it trying to find `Markdig` here in the `__WASM__` guard for Mac, iOS, and Android?

Arlodotexe: We manually define these in `./common/Labs.ProjectIdentifiers.props`. It may be that we have a bad condition in that file, causing `IsWasm` to be set to `true` under Droid, macOS and iOS.

platforms/CommunityToolkit.Labs.Wasm/CommunityToolkit.Labs.Wasm.csproj
```diff
@@ -34,4 +34,8 @@
     <EmbeddedResource Include="WasmCSS\Fonts.css" />
     <EmbeddedResource Include="WasmScripts\AppManifest.js" />
   </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Markdig" Version="0.30.4" />
```
michael-hawker, line 39: Oh, I have to add the Markdig reference to all WASM sample heads... 🤔

I need to do this in `Labs.Heads.Wasm.props` instead, oops. That's what I get for just adding the package via VS... 🙈

common/CommunityToolkit.Labs.Shared/Renderers/Markdown/MarkdownTextBlock.cs
michael-hawker: @Arlodotexe looks like this just may be a VS issue. CI seemed fine. Not sure what that means for building those heads locally though.

michael-hawker:
Looks like we can get #211 in, so we'll merge that and then I'll update this PR on top of it.

michael-hawker:
@Arlodotexe I've updated this and tested locally on UWP, WASDK, and WASM. Working great there. Weird build issues with other platform heads trying to build the `__WASM__` section instead of the uno section though...

Let's see what happens in the CI. If it's clean, we should merge to get the improvements here. And then file a separate issue to figure out what's going on with VS?

michael-hawker:
@niels9001 icons looks great now on WASM!

image

common/CommunityToolkit.Labs.Core.SourceGenerators/ToolkitSampleMetadataGenerator.Documentation.cs
```diff
@@ -29,6 +29,12 @@ public partial class ToolkitSampleMetadataGenerator
     private const string FrontMatterRegexSubcategoryExpression = @"^subcategory:\s*(?<subcategory>.*)$";
     private static readonly Regex FrontMatterRegexSubcategory = new Regex(FrontMatterRegexSubcategoryExpression, RegexOptions.Compiled | RegexOptions.IgnoreCase | RegexOptions.Multiline);
 
+    private const string FrontMatterRegexDiscussionIdExpression = @"^labs-discussion:\s*(?<discussionId>.*)$";
```
Arlodotexe, line 32: Let's remove "labs" from these two new fields. This source generator will serve both Labs and the main Toolkit.

common/CommunityToolkit.Labs.Shared/Renderers/ToolkitDocumentationRenderer.xaml.cs
```diff
@@ -213,4 +213,12 @@ private async void MarkdownTextBlock_LinkClicked(object sender, LinkClickedEvent
         }
     }
 #endif
+
+    public static Uri ToLabsUri(string path, int id) => new Uri($"https://github.com/CommunityToolkit/Labs-Windows/{path}/{id}");
```
Arlodotexe, line 217: Let's remove "Labs" from this as well, for the same reason as above.

Arlodotexe: Also, can we pull this from an environment variable? This can't stay hardcoded to the Labs repo.

Arlodotexe (changes requested):

Arlodotexe (changes requested):
Caught a few issues that'll clash with the main toolkit down the road.

common/CommunityToolkit.Labs.Shared/Renderers/ToolkitDocumentationRenderer.xaml.cs
michael-hawker: I'll see if I can pull from a prop in the csproj or something maybe? Or maybe a config file in the root that the sample heads reference? I'll investigate.

michael-hawker: I've cleaned-up the YAML front-matter label names. I'll open a 2nd PR for the url and the metadata reading as that's more complex (but close on the approach there, just need to debug some WASM stuff for Uno, but working in UWP/WinAppSDK now).

Arlodotexe (approved):

PR #317, opened 2022-11-17
[SettingsControls] Narrator improvements

niels9001:
Some a11y improvements to unblock the Photos team (Narrator was not announcing everything it should).

I also saw that SettingsExpander.Content was of type UIElement, but we use _object_ for SettingsCard so changed that as well.

### DIFF
--- a/labs/SettingsControls/src/SettingsCard/SettingsCard.cs
+++ b/labs/SettingsControls/src/SettingsCard/SettingsCard.cs
@@ -49,18 +49,13 @@ public partial class SettingsCard : ButtonBase
 
     private void RegisterAutomation()
     {
-        if (Header != null && Header.GetType() == typeof(string))
+        if (Header is string headerString && headerString != string.Empty)
         {
-            string? headerString = Header.ToString();
-            if (!string.IsNullOrEmpty(headerString))
+            AutomationProperties.SetName(this, headerString);
+            // We don't want to override an AutomationProperties.Name that is manually set, or if the Content basetype is of type ButtonBase (the ButtonBase.Content will be used then)
+            if (Content is UIElement element && string.IsNullOrEmpty(AutomationProperties.GetName(element)) && element.GetType().BaseType != typeof(ButtonBase))
             {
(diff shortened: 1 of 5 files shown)


labs/SettingsControls/src/SettingsCard/SettingsCard.cs
```diff
@@ -55,12 +55,9 @@ private void RegisterAutomation()
...
             {
                 AutomationProperties.SetName(this, headerString);
-            }
 
-            if (Content != null && Content.GetType() != typeof(Button))
-            {
-                // We do not want to override the default AutomationProperties.Name of a button. Its Content property already describes what it does.
-                if (!string.IsNullOrEmpty(headerString))
+                // We don't want to override an AutomationProperties.Name that is manually set, or if the Content basetype is of type ButtonBase (the ButtonBase.Content will be used then)
+                if (Content != null && string.IsNullOrEmpty(AutomationProperties.GetName((UIElement)Content)) && Content.GetType().BaseType != typeof(ButtonBase))
```
michael-hawker, line 60: since you've changed type to `object` now, you need to do additional check of: `Content is UIElement element` and then use `element.Content`.

labs/SettingsControls/src/SettingsExpander/SettingsExpander.cs
```diff
@@ -30,21 +30,11 @@ private void RegisterAutomation()
         if (Header != null && Header.GetType() == typeof(string))
```
michael-hawker, line 30:
```
        if (Header is string headerString && headerString != string.Empty)
```
Can simplify using the C# `is` syntax. See a few places around here that could benefit from this.

labs/SettingsControls/src/SettingsCard/SettingsCard.cs
michael-hawker: Should also just be able to do `Content is not ButtonBase`

Arlodotexe (approved):

PR #327, opened 2022-12-02
Generate AssemblyInfo from msbuild vs. including in all projects as file

### LINKED ISSUE #326: Infra: "ToSourceUri" is hardcoded to labs
## Background
As part of [#308](https://github.com/CommunityToolkit/Labs-Windows/pull/308/files#r1037688649), we added the ability to navigate directly to a GitHub issue or Discussion from the UI.

## Problem
The repo URL and Labs moniker were hardcoded into the UI. Before we can take the Labs infrastructure to the main toolkit, this value will need to be pulled from somewhere outside of the `/common/` folder.

michael-hawker:
Fixes #326 

This approach removes another file that needs to be duplicated across experiments.

Instead of having hard-coded `Propertes\AssemblyInfo.cs` files, we generate the source for those via an msbuild task, `WriteCodeFragment` using `AssemblyAttributes`.

We do this with the msbuild task and populating the values before the build is setup with a custom target. We can then use reflection at runtime to get our custom assembly attribute which now stores the `PackageProjectUrl` from the root common props.

This will let us store various metadata needed for configuration of our sample app that can work across repositories when we use this infrastructure for the main toolkit.

## TODO

- [x] Investigate: WASM not reading AssemblyMetadata (was generating fine, but binding was early due to https://github.com/unoplatform/uno/issues/2666 https://github.com/unoplatform/uno/issues/2872 - moved to constructor)
- [x] Replace `$projectname$` type values with the props from `Windows.Toolkit.Common.props` instead directly?
- [ ] ~~Use this approach for `InternalVisibleTo` and remove `AdditionalAssemblyInfo` as well for library projects?~~ Opened #329 for future work tracking
- [x] Build outputs new WinUI 3 errors about min Platform Version? (at least locally, will check CI) (Edit: checked CI has the same warnings as the last main build)


michael-hawker:
Hmmm... Previous Metadata generation task worked fine on WASM in the CI, but the new fix to move it to `obj` folder failed:

```
/home/runner/work/Labs-Windows/Labs-Windows/common/Labs.Head.props(136,5): error MSB3713: The file "obj/Debug/net5.0/AssemblyInfo.g.cs" could not be created. Could not find a part of the path '/home/runner/work/Labs-Windows/Labs-Windows/platforms/CommunityToolkit.Labs.Wasm/obj/Debug/net5.0/AssemblyInfo.g.cs'. [/home/runner/work/Labs-Windows/Labs-Windows/platforms/CommunityToolkit.Labs.Wasm/CommunityToolkit.Labs.Wasm.csproj]
```

This is odd, as I tested on UWP, WinAppSDK, and WASM locally first.

common/Labs.Head.props
```diff
@@ -59,5 +59,80 @@
...
+      </AssemblyAttributes>
+      <AssemblyAttributes Include="CommunityToolkit.Attributes.PackageProjectUrlAttribute">
+        <_Parameter1>$(PackageProjectUrl)</_Parameter1>
+      </AssemblyAttributes>
+    </ItemGroup>
+
+    <!-- Extra attributes specific to platforms -->
+    <ItemGroup Condition="'$(IsWinAppSdk)' == 'true'">
+      <AssemblyAttributes Include="System.Runtime.Versioning.TargetPlatformAttribute">
+        <_Parameter1>Windows10.0.19041.0</_Parameter1><!-- TODO: Grab from variable -->
```
Arlodotexe, line 124: The property values for our TargetFrameworks are always loaded before `Labs.Head.props` is imported.

You should have those values here, let's cross these 2 TODOs off in this PR.

labs/CanvasLayout/samples/CanvasLayout.Uwp/CanvasLayout.Uwp.csproj
```diff
@@ -35,9 +35,6 @@
     <AssemblyName>CanvasLayoutExperiment.Samples.Uwp</AssemblyName>
     <ProjectGuid>{C3F460EC-A0A6-4B22-A91A-6CF98B30FD8A}</ProjectGuid>
   </PropertyGroup>
-  <ItemGroup>
-    <Compile Include="Properties\AssemblyInfo.cs" />
```
Arlodotexe, line 39: Need to file a tech debt ticket for this just before we close off. New experiments still in PR will need to get this change.

common/CommunityToolkit.Labs.Shared/Helpers/PackageProjectUrlAttribute.cs
```diff
@@ -0,0 +1,18 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace CommunityToolkit.Attributes;
```
Arlodotexe, line 7: Namespace doesn't match folder structure. We're likely to have more attributes like this in the near future, let's move this to a dedicated `Attributes` folder.

Arlodotexe (changes requested):

common/Labs.Head.props
michael-hawker: Oh, oops, I knew there was another one. I grabbed the ones for the Company/Copyright, and missed this one below. Nice catch!

labs/CanvasLayout/samples/CanvasLayout.Uwp/CanvasLayout.Uwp.csproj
michael-hawker: Yeah, if they update they'll get duplicate metadata things, so should be pretty obvious.

common/CommunityToolkit.Labs.Shared/Helpers/PackageProjectUrlAttribute.cs
michael-hawker: Yeah, was trying to avoid making a folder for one file... 😋

common/Labs.Head.props
michael-hawker: @Arlodotexe we need to centralize the min/target versions better as this value isn't the same as the Target Framework one (`net6.0-windows10.0.19041.0`) albeit similar.

I suggest we create a follow-up issue.

michael-hawker: This is already tracked by https://github.com/CommunityToolkit/Labs-Windows/issues/145, added a comment there.

common/CommunityToolkit.Labs.Shared/Renderers/ToolkitDocumentationRenderer.xaml.cs
```diff
@@ -24,9 +24,18 @@ public sealed partial class ToolkitDocumentationRenderer : Page
...
 
     public ToolkitDocumentationRenderer()
     {
+        // Check and Cache here before we use in XAML.
+        if (ProjectUrl == null)
+        {
+            ProjectUrl = Assembly.GetExecutingAssembly()?.GetCustomAttribute<CommunityToolkit.Attributes.PackageProjectUrlAttribute>()?.PackageProjectUrl;
+        }
+
+        // TODO: If ProjectUrl is null should we log an error?
```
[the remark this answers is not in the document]
michael-hawker: image

Buttons will now just not show-up if the url is missing. Was a bit of a pain as the x:Bind's were still trying to create the Uri and throwing an exception even with x:Load...?

Also added a MSBuild Warning when it's not set (though I don't see it in the VS build output?).

Arlodotexe (approved):

labs/CanvasLayout/samples/CanvasLayout.Uwp/CanvasLayout.Uwp.csproj
Arlodotexe: Looks like there's only 2 open experiment PRs. Now that this is merged to main, I'll update those branches to latest main and fix this, before they do.

PR #333, opened 2022-12-08
[SettingsExpander] Adding ItemsHeader/ItemsFooter property

niels9001:
This PR introduces a ItemsHeader/ItemsFooter property, a way to host custom content add the top or bottom of a SettingsExpander. This is a common practice across W11 Settings to e.g. visualize links or other messages.

```xml
<labs:SettingsExpander.ItemsFooter>
     <muxc:InfoBar Title="This is the footer">
          <muxc:InfoBar.ActionButton>
               <HyperlinkButton Content="It can host custom content" />
          </muxc:InfoBar.ActionButton>
     </muxc:InfoBar>
</labs:SettingsExpander.ItemsFooter>
```

Result:
Footer2

### DIFF
--- a/labs/SettingsControls/src/SettingsExpander/SettingsExpander.Properties.cs
+++ b/labs/SettingsControls/src/SettingsExpander/SettingsExpander.Properties.cs
@@ -45,6 +45,15 @@ public partial class SettingsExpander
         new PropertyMetadata(defaultValue: null));
 
     /// <summary>
+    /// The backing <see cref="DependencyProperty"/> for the <see cref="Content"/> property.
+    /// </summary>
+    public static readonly DependencyProperty FooterProperty = DependencyProperty.Register(
+        nameof(Footer),
+        typeof(object),
+        typeof(SettingsExpander),
+        new PropertyMetadata(defaultValue: null));
+
+    /// <summary>
     /// The backing <see cref="DependencyProperty"/> for the <see cref="IsExpanded"/> property.
     /// </summary>
     public static readonly DependencyProperty IsExpandedProperty = DependencyProperty.Register(
@@ -91,6 +100,15 @@ public partial class SettingsExpander
     {
         get => (object)GetValue(ContentProperty);
         set => SetValue(ContentProperty, value);
+    }
+
+    /// <summary>
+    /// Gets or sets the Footer.
+    /// </summary>
+    public object Footer
+    {
+        get => (object)GetValue(FooterProperty);
+        set => SetValue(FooterProperty, value);
     }
 
     /// <summary>
--- a/labs/SettingsControls/src/SettingsExpander/SettingsExpander.Properties.cs
+++ b/labs/SettingsControls/src/SettingsExpander/SettingsExpander.Properties.cs
@@ -49,7 +49,7 @@ public partial class SettingsExpander
     /// </summary>
     public static readonly DependencyProperty FooterProperty = DependencyProperty.Register(
         nameof(Footer),
-        typeof(object),
+        typeof(UIElement),
         typeof(SettingsExpander),
         new PropertyMetadata(defaultValue: null));
 
@@ -105,9 +105,9 @@ public partial class SettingsExpander
     /// <summary>
     /// Gets or sets the Footer.
     /// </summary>
-    public object Footer
+    public UIElement Footer
     {
-        get => (object)GetValue(FooterProperty);
+        get => (UIElement)GetValue(FooterProperty);
         set => SetValue(FooterProperty, value);
     }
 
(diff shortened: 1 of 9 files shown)


labs/SettingsControls/src/SettingsExpander/SettingsExpander.Properties.cs
```diff
@@ -93,6 +102,15 @@ public object Content
         set => SetValue(ContentProperty, value);
     }
 
+    /// <summary>
+    /// Gets or sets the Footer.
+    /// </summary>
+    public object Footer
```
michael-hawker, line 108: Should this be a `UIElement`? I think the guidance would be if you can provide a template than use `object` (i.e. if we had a `FooterTemplate` property as well), otherwise we use `UIElement` to say that it should be some other XAML construct, eh?

Wondering if we need to do the same for `Header`? Or since we're using `ContentPresenter` if we also want to add `HeaderTemplate`/`FooterTemplate` properties?

niels9001: As discussed, I'll move this to UIElement

michael-hawker: Don't forget about the same for `Header` as well 😉

niels9001: Ah.. looked at that, and then I realized we use the Header and Description are bind the SettingsCard - where we use these properties (both of type object for accessibility purposes (i.e. checking if these are of type string and then use those as the AutomationProperties.Name of the control that is set for the Content).

Should we leave those as is (for now)?

labs/SettingsControls/src/SettingsExpander/SettingsExpander.xaml
```diff
@@ -89,7 +89,10 @@
                                                    IsClickEnabled="False" />
                             </muxc:Expander.Header>
                             <muxc:Expander.Content>
-                                <ItemsPresenter TabFocusNavigation="Local" />
+                                <StackPanel>
+                                    <ItemsPresenter TabFocusNavigation="Local" />
```
michael-hawker, line 93: Wonder if [ItemsPresenter.Footer](https://learn.microsoft.com/uwp/api/windows.ui.xaml.controls.itemspresenter.footer) would be better here? (or at least simplify)

michael-hawker: Discussing more let's call these properties `ItemsFooter` and add an `ItemsHeader` and then bind them to the ItemsPresenter (assuming that works). Gives us the most flexibility for future expansion.

We can later use `Footer` to be an attached item to the Expander outside of the expander itself.

niels9001: Done!

michael-hawker:
This is so cool! And so simple in the end! Will pull this down to try out locally :)

I added Uno issue https://github.com/unoplatform/uno/issues/1979 to our tracking board for the support for ItemsPresenter Header/Footer.

michael-hawker:
@niels9001 wasn't sure if it was the Windows App SDK at first:

image

But at least it's consistent on UWP:

image

Seems like there's a bit of extra padding around the InfoBar? (The InfoBar isn't rounded on the top as well, think that's their style?)

Very nit-picky, and we can probably follow-up separately (with custom InfoBar styling injection or padding fixes); so will merge for now, as seems great!

michael-hawker (approved):
I did have a weird issue where the UWP sample app crashed when opening the Expander on the ItemsSource example page when it wasn't attached to a debugger... but don't think that's specific to this PR. FYI @Arlodotexe 

Didn't get much in Event Viewer beyond System32\KERNELBASE.dll and combase.dll

michael-hawker:
Uno created a new tracking issue specifically for ItemsPresenter here: https://github.com/unoplatform/uno/issues/10745 - I've updated our tracking board to point to that one vs. the ListViewBase one.

niels9001:
> @niels9001 wasn't sure if it was the Windows App SDK at first:
> 
> image
> 
> But at least it's consistent on UWP:
> 
> image
> 
> Seems like there's a bit of extra padding around the InfoBar? (The InfoBar isn't rounded on the top as well, think that's their style?) Using Screen Ruler you'll see there's a pixel difference, two with 150% DPI - between the app's left border and the infobar/cards:
> 
> image
> 
> image
> 
> (There's also an extra pixel gap under the InfoBar, both Header and Footer. The footer one is darker gutter, so it's harder to notice.)
> 
> Very nit-picky, and we can probably follow-up separately (with custom InfoBar styling injection or padding fixes); so will merge for now, as seems great!

Ah, good catch! There's a 1px Border around the InfoBar. For the sample, I'll remove it and also add the correct CornerRadius for the InfoBar used in the header!

image

Issue #207, opened 2022-07-19
Add more Sample Option Attributes


michael-hawker:
Related #149 

Realized as we've been building more samples, that just having `bool` and `multi-choice` as our options is a bit limiting, so wanted to list out some other simple scenarios for `ToolkitSample*OptionAttribute`s we may want in the future:

| Name | Parameters | Description |
|--------|---------------|----------------------|
| `ToolkitSampleCommandOptionAttribute` | `label` + `nameof(SomeMethodOnSampleClass)` | Creates a `Button` with `label` content which will execute the command when clicked |
| `ToolkitSampleTextOptionAttribute` | `title` (as normal) + `placeholder` | Creates a `TextBox` with `placeholder` text, allows user to type freeform text |
| `ToolkitSampleSliderOptionAttribute` | `min` + `max` + `step` + `initial` | Creates a `Slider` |

We may also want a flag for MultiChoice to change whether it's radio buttons or a combobox?

We can always refer to [our existing sample methods for ideas](https://github.com/CommunityToolkit/WindowsCommunityToolkit/tree/main/Microsoft.Toolkit.Uwp.SampleApp#2-binding-text).

michael-hawker:
Also found an issue trying to TwoWay bind to a `ToolkitSampleBoolOption` value. Uno throws an error in its XAML source generator that it can't find the property.

Arlodotexe:
> Also found an issue trying to TwoWay bind to a `ToolkitSampleBoolOption` value. Uno throws an error in its XAML source generator that it can't find the property.

Thanks, we'll get that fixed up when we move focus back to Labs again.

PR #336, opened 2022-12-12
More Sample Option Attributes

niels9001:
Related to #207

- Adding the TextOption (PlaceholderText can be set)
- Adding the NumericOption (Initial, Min, Max, Step, ShowAsNumberBox can be set)
- Replacing the RadioButtons in the MultiChoiceOption with ComboBoxes to save space.

Settings

How to test:
- Run ProjectTemplate.sln, or:

```xml
    <TextBlock
                Text="{x:Bind TextContent, Mode=OneWay}"
                FontFamily="{x:Bind TextFontFamily, Mode=OneWay}"
                Foreground="{x:Bind TextForeground, Mode=OneWay}"
                Visibility="{x:Bind IsTextVisible, Mode=OneWay}"
                FontSize="{x:Bind TextSize, Mode=OneWay}" />
```
- Add these options:
```
[ToolkitSampleTextOption("TextContent", "This is text", Title = "Input the text")]
[ToolkitSampleMultiChoiceOption("TextFontFamily", "Segoe UI", "Arial", "Consolas", Title = "Font family")]
[ToolkitSampleMultiChoiceOption("TextForeground", "Teal : #0ddc8c", "Sand : #e7a676", "Dull green : #5d7577", Title = "Text foreground")]
[ToolkitSampleBoolOption("IsTextVisible", true, Title = "IsVisible")]
[ToolkitSampleNumericOption("TextSize", 12, 8, 48, 2, false, Title = "FontSize")]  // To show as Slider (default)
[ToolkitSampleNumericOption("TextSize", 12, 8, 48, 2, true, Title = "FontSize")]  // To show as NumberBox
```


michael-hawker:
@niels9001 added link to #207 at top of your description. 

FYI @Arlodotexe as well.

I believe the optional `title` parameter is to give it a different header vs. just using the name of the binding property. However, we have this inconsistency between putting it as the 2nd parameter or at the end as optional.

I feel like we maybe want to be more consistent here? Or should we just get rid of `title` all together and either just use the same name as the variable (that's going to be shown in the XAML anyway for the example code) or just auto-insert a space before each capital letter?

Thoughts?

common/CommunityToolkit.Labs.Core.SourceGenerators/Attributes/ToolkitSampleTextOptionAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+        PlaceholderText = placeholderText;
+    }
+
+    /// <summary>
+    /// The source generator-friendly type name used for casting.
+    /// </summary>
+    internal override string TypeName { get; } = "string";
+
+    /// <summary>
+    /// A label to display along the boolean option.
```
Arlodotexe, line 35: Incorrect summary for PlaceholderText.

Arlodotexe:
> I feel like we maybe want to be more consistent here? Or should we just get rid of title all together and either just use the same name as the variable (that's going to be shown in the XAML anyway for the example code) or just auto-insert a space before each capital letter?

@michael-hawker That would make it match what the current WCT Sample App does 👍

michael-hawker:
> Custom titles should be optional (but enabled, for multi-control samples)

@Arlodotexe what do you mean? What would a 'multi-control' sample be? Like where would we need to customize this.

I think it's clearer if we have the text matching the property they're going to see bound in the XAML example code, right? Otherwise if you're looking at the sample code you don't know what the toggle/slider/etc... is manipulating.

Arlodotexe:
What I mean is that there will be (currently are, in the case of Rive) cases where a label might not be needed or where a custom description provides more value than using the variable name.

 We shouldn't lock ourselves out of having this option. Using the `bindingName` as the default title is a great idea, though.

common/CommunityToolkit.Labs.Core.SourceGenerators/Attributes/ToolkitSampleSliderOptionAttribute.cs
```diff
@@ -0,0 +1,57 @@
...
+/// <para/>
+/// </remarks>
+[AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+public sealed class ToolkitSampleSliderOptionAttribute : ToolkitSampleOptionBaseAttribute
+{
+    /// <summary>
+    /// Creates a new instance of <see cref="ToolkitSampleSliderOptionAttribute"/>.
+    /// </summary>
+    /// <param name="bindingName">The name of the generated property, which you can bind to in XAML.</param>
+    /// <param name="choices">A list of the choices to display to the user. Can be literal values, or labeled values. Use a " : " separator (single colon surrounded by at least 1 whitespace) to separate a label from a value.</param>
```
Arlodotexe, line 22: These parameter descriptions will need to be updated.

common/CommunityToolkit.Labs.Core.SourceGenerators/Attributes/ToolkitSampleSliderOptionAttribute.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// </summary>
+    public double? Min { get; }
+
+    /// <summary>
+    /// The maximum value
+    /// </summary>
+    public double? Max { get; }
+
+    /// <summary>
+    /// The step value
```
Arlodotexe, line 49:
```suggestion
    /// The step value.
```

common/CommunityToolkit.Labs.Core.SourceGenerators/Attributes/ToolkitSampleSliderOptionAttribute.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// </summary>
+    public double? Initial { get; }
+
+    /// <summary>
+    /// The minimal value
+    /// </summary>
+    public double? Min { get; }
+
+    /// <summary>
+    /// The maximum value
```
Arlodotexe, line 44:
```suggestion
    /// The maximum value.
```

common/CommunityToolkit.Labs.Core.SourceGenerators/Attributes/ToolkitSampleSliderOptionAttribute.cs
```diff
@@ -0,0 +1,57 @@
...
+        Step = step;
+    }
+
+    /// <summary>
+    /// The default start value
+    /// </summary>
+    public double? Initial { get; }
+
+    /// <summary>
+    /// The minimal value
```
Arlodotexe, line 39:
```suggestion
    /// The minimal value.
```

common/CommunityToolkit.Labs.Core.SourceGenerators/Attributes/ToolkitSampleSliderOptionAttribute.cs
```diff
@@ -0,0 +1,57 @@
...
+        : base(bindingName, null, title)
+    {
+        Initial = initial;
+        Min = min;
+        Max = max;
+        Step = step;
+    }
+
+    /// <summary>
+    /// The default start value
```
Arlodotexe, line 34:
```suggestion
    /// The default start value.
```

michael-hawker:
Discussed in our team meeting. We'll set `Title` to be a settable property:

https://github.com/CommunityToolkit/Labs-Windows/blob/084e2db957cb734b544f1db014c48d2ba4ab5fb6/common/CommunityToolkit.Labs.Core.SourceGenerators/Attributes/ToolkitSampleOptionBaseAttribute.cs#L38

Remove from constructor here:

https://github.com/CommunityToolkit/Labs-Windows/blob/084e2db957cb734b544f1db014c48d2ba4ab5fb6/common/CommunityToolkit.Labs.Core.SourceGenerators/Attributes/ToolkitSampleOptionBaseAttribute.cs#L18

This lets us remove Title from other attributes like so:

https://github.com/CommunityToolkit/Labs-Windows/blob/084e2db957cb734b544f1db014c48d2ba4ab5fb6/common/CommunityToolkit.Labs.Core.SourceGenerators/Attributes/ToolkitSampleMultiChoiceOptionAttribute.cs#L24

And then we should be able to initialize them like so:

```cs
[ToolkitSampleBoolOption("IsTextVisible", true, Title = "IsVisible")]
[ToolkitSampleSliderOption("TextSize", 50, 1, 200, 10, Title = "FontSize")]
[ToolkitSampleTextOption("TextText", "Placeholder text", Title = "Input the text")]
[ToolkitSampleMultiChoiceOption("TextFontFamily", "Segoe UI", "Arial", "Consolas", Title = "Font family")]
```

common/CommunityToolkit.Labs.Core.SourceGenerators/Metadata/ToolkitSampleTextOptionMetadataViewModel.cs
```diff
@@ -0,0 +1,81 @@
...
+/// </remarks>
+public class ToolkitSampleTextOptionMetadataViewModel : IGeneratedToolkitSampleOptionViewModel
+{
+    private string? _title;
+    private object _value;
+
+    /// <summary>
+    /// Creates a new instance of <see cref="ToolkitSampleTextOptionAttribute"/>.
+    /// </summary>
+    public ToolkitSampleTextOptionMetadataViewModel(string id, string placeholderText = null, string? title = null)
```
Arlodotexe, line 24:
```suggestion
    public ToolkitSampleTextOptionMetadataViewModel(string id, string placeholderText? = null, string? title = null)
```

niels9001:
@Arlodotexe CI complains about 'Arguments missing'.. but I can't figure out where that'd be.. Do you see if I am missing something obvious :)?

niels9001:
> @Arlodotexe CI complains about 'Arguments missing'.. but I can't figure out where that'd be.. Do you see if I am missing something obvious :)?

Fixed! Thanks for the help.

@michael-hawker Added a bool to switch between NumberBox / Slider as well - see original post. Once this passes CI, this should be ready to review :)!

niels9001:
One thing that's not working: the titles do not show up.

michael-hawker:
@niels9001 the build is failing on the tests that check/look that these attributes are parsed/working as expected, so you'll need to update/add to those as well.

As for the Title not displaying, @Arlodotexe thoughts?

common/CommunityToolkit.Labs.Shared/Renderers/GeneratedSampleOptionsRenderer.xaml
```diff
@@ -16,28 +16,54 @@
...
             <local:GeneratedSampleOptionTemplateSelector.MultiChoiceOptionTemplate>
                 <DataTemplate x:DataType="metadata:ToolkitSampleMultiChoiceOptionMetadataViewModel">
-                    <StackPanel>
-                        <TextBlock Text="{x:Bind Title}"
-                                   Visibility="{x:Bind local:GeneratedSampleOptionsRenderer.NullOrWhiteSpaceToVisibility(Title)}" />
-                        <muxc:RadioButtons ItemsSource="{x:Bind Options}"
-                                           SelectedIndex="0"
-                                           SelectedItem="{x:Bind Value, Mode=TwoWay}" />
-                    </StackPanel>
+                    <ComboBox Header="{x:Bind Title, Mode=OneWay}"
```
michael-hawker, line 25: @niels9001 does this one work? I noticed the other `Title` are only `OneTime`, is that a difference?

niels9001: Unfortunately not :(

michael-hawker:
image

Have a fix incoming for the issue with Title not appearing.

Since we switched to using the property on the Attribute, this wasn't setup when we rehydrated the attributes to store for creating the sample registry, thus the Title was blank and not propagated to the UI.

We needed an extra step to grab the property value on the attribute in the source generator when it ran so that we could add it back into the attribute we use for storage to generate the registry from.

This would then let it flow up back the rest of the chain to the UI and display properly once fixed.

Had to also add in some extra helper to the tests for checking the generated source (we can probably do more to clean this up later), would be nice to compile the source or get the compiled instance of the source to check actual values via code vs. comparing strings. Would make it easier to call out specific things missing too vs. having to check manually if a test fails.

michael-hawker (approved):

PR #348, opened 2023-01-05
🧪 [Experiment] TokenView

niels9001:
This PR introduces the TokenView control. More information about this control here: https://github.com/CommunityToolkit/Labs-Windows/discussions/323

A TokenView is used to display TokenItems using Fluent Design. It's derived of ListViewBase to handle most properties but has some unique behavior to make it easy for developers to adopt this UI patterns.

### Using the TokenView

The TokenView can be used with the following XAML (Binding is supported as well)

```xml
 <TokenView>
     <TokenItem Content="All" />
     <TokenItem Content="Work" Icon="{ui:FontIcon Glyph=&#xECAA;}"/>
     <TokenItem Content="Apps" Icon="{ui:FontIcon Glyph=&#xE821;}"/>
     <TokenItem Content="Documents" Icon="{ui:FontIcon Glyph=&#xE821;}"/>
     <TokenItem Content="Web" Icon="{ui:FontIcon Glyph=&#xE12B;}"/>
     <TokenItem Content="Settings" Icon="{ui:FontIcon Glyph=&#xE115;}"/>
     <TokenItem Content="People" Icon="{ui:FontIcon Glyph=&#xE13D;}"/>
     <TokenItem Content="Folders" Icon="{ui:FontIcon Glyph=&#xE8B7;}"/>
     <TokenItem Content="Photos" Icon="{ui:FontIcon Glyph=&#xE91B;}"/>
</TokenView>
```
### Orientation & selection
The `Orientation` can be set to `Horizontal` (default) or `Wrapped`. The horizontal mode will automatically add scrollbuttons to scroll through the list of items. There's also touch support to scroll through the list. The wrapped state will show all items stacked. 

Both single and multiselect is supported.

TokenViewSample1

### Removing TokenItems
The `IsRemoveable` property of a `TokenItem` can be set - a remove button will appear. The `CanRemoveTokens` can be set on the `TokenView` to set this for all children. The scrollbuttons will automatically (disappear) based on the amount of Tokens and the scrollviewer offset.

TokenViewSample2

### Attribution
TokenView uses code created for the [deprecated the Community Toolkit's TabView control](https://github.com/CommunityToolkit/WindowsCommunityToolkit/tree/5f4caca50c99417d628b445d722fbdb7749d5b1c/Microsoft.Toolkit.Uwp.UI.Controls/TabView).


michael-hawker:
Looks like an issue with the visual tree helpers:

```
 D:\a\Labs-Windows\Labs-Windows\components\CanvasLayout\tests\ExampleCanvasLayoutTestClass.cs(59,30): error CS0121: The call is ambiguous between the following methods or properties: 'CommunityToolkit.Labs.WinUI.VisualTree.FindDescendant<T>(Windows.UI.Xaml.DependencyObject)' and 'Microsoft.Toolkit.Uwp.UI.DependencyObjectExtensions.FindDescendant<T>(Windows.UI.Xaml.DependencyObject)' [D:\a\Labs-Windows\Labs-Windows\common\ProjectHeads\AllComponents\Tests.Uwp\CommunityToolkit.Tests.Uwp.csproj]
 ```

Not sure if this was a before thing or after thing...

**Comment by niels9001** [comment]
> Looks like an issue with the visual tree helpers:
> 
> ```
>  D:\a\Labs-Windows\Labs-Windows\components\CanvasLayout\tests\ExampleCanvasLayoutTestClass.cs(59,30): error CS0121: The call is ambiguous between the following methods or properties: 'CommunityToolkit.Labs.WinUI.VisualTree.FindDescendant<T>(Windows.UI.Xaml.DependencyObject)' and 'Microsoft.Toolkit.Uwp.UI.DependencyObjectExtensions.FindDescendant<T>(Windows.UI.Xaml.DependencyObject)' [D:\a\Labs-Windows\Labs-Windows\common\ProjectHeads\AllComponents\Tests.Uwp\CommunityToolkit.Tests.Uwp.csproj]
> ```
> 
> Not sure if this was a before migration thing or after thing...

Fixed! This should be ready for a first v0.1 review :)

**Comment by michael-hawker** [comment]
I'll rebase this this morning.

@niels9001 we just need confirmation if we have the updates selection workaround (and comment) that we used in #357?

**Comment by michael-hawker** [comment]
@niels9001 should be rebased now, found a piece of feedback 😋

File: components/TokenView/src/TokenView/TokenViewOrientation.cs [hunk 1]
```diff
@@ -0,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace CommunityToolkit.Labs.WinUI;
+
+public enum TokenViewOrientation
```
michael-hawker, line 7: Separated out this enum as otherwise it's in a weird namespace.

However, I'm not a big fan of this property... Do we see more than two options here? Either we could have a boolean like `IsWrapped` or call it `WrappingMode` instead with like `NoWrap` and `Wrap` (like `TextWrapping` on `RichTextBlock`)?

michael-hawker:
Oops, forgot to check test file, re-running with fix.

components/TokenView/samples/TokenViewItemsSourceSample.xaml
```diff
@@ -0,0 +1,35 @@
...
+                        SelectionChanged="TokenView_SelectionChanged"
+                        SelectionMode="Single">
+            <labs:TokenView.ItemTemplate>
+                <DataTemplate x:DataType="local:MyDataModel">
+                    <Grid>
+                        <TextBlock Margin="16,0,16,0"
+                                   Style="{StaticResource CaptionTextBlockStyle}"
+                                   Text="{x:Bind Name}" />
+                    </Grid>
+                </DataTemplate>
```
michael-hawker, line 25: Just wondering if we want this to work like the new-style where you specify the item? That way you can bind to content and/or icon? Otherwise you have to re-style the icon yourself here in this scenario?

That's a bigger change though and as we know has complications potentially. So maybe something to think about for next big iteration from any other feedback we get?

components/TokenView/src/TokenView/TokenViewOrientation.cs
michael-hawker, line 1: This can be removed now, eh?

michael-hawker (commented):
Need to take it for a spin some time, just left a few code comments for future.

One small clean-up with a extra file I think.

@Arlodotexe was building locally, but good to squash this into `main` and continue improving as we move forward with feedback.

components/TokenView/src/TokenView/TokenViewOrientation.cs
michael-hawker: Removed

components/TokenView/src/TokenView/TokenView.Events.cs
```diff
@@ -0,0 +1,160 @@
...
+            }
+            else
+            {
+                if (_tokenViewScroller != null)
+                {
+                    _tokenViewScroller.UpdateLayout();
+                }
+                Items.Remove(item);
+            }
+        }
```
michael-hawker, line 123: Found a bug, you also need to check if it's the SelectedItem, and if so set that to null:

image

image

components/TokenView/src/TokenView/TokenView.Events.cs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    private void Token_Removing(object? sender, TokenItemRemovingEventArgs e)
+    {
+        if (ItemFromContainer(e.TokenItem) is object item)
+        {
+            var args = new TokenItemRemovingEventArgs(item, e.TokenItem);
+            TokenItemRemoving?.Invoke(this, args);
+
+            if (ItemsSource != null)
```
michael-hawker, line 111: Do we want to provide a hook to allow for the `Handled` pattern and prevent a token from being removed based on end-dev's code? We did this in the TokenizingTextBox with the deferred events (so they could use async with a deferral to pop-up a dialog for instance).

Maybe an issue we should log for future work? (Especially once we've ported those helpers to be in the common namespace.)

michael-hawker (approved):
Almost working in WASM... some weird issue with the Data Template for the binding sample:
image

Wrapped mode works, but unwrapped the icon is missing in the button:
image

And the button doesn't move to the left-side to allow going back to the left:
image

Everything else seems to work great.

Since it compiles, we should probably update Uno and see if that fixes things before we investigate further.

I'll fix the SelectedItem issue quick here, as I think that's the only blocking issue.

components/TokenView/src/TokenView/TokenView.Events.cs
michael-hawker: Fixed in https://github.com/CommunityToolkit/Labs-Windows/pull/348/commits/877f39393320bf142118900b24a7b5b9673b62de

michael-hawker:
@Arlodotexe I fixed the sample selection issue, everything else from my review is a follow-on task we can file/follow-up on later. Think we're good to go once you build locally and approve. 🎉🎉🎉

Arlodotexe:
> @Arlodotexe I fixed the sample selection issue, everything else from my review is a follow-on task we can file/follow-up on later. Think we're good to go once you build locally and approve. 🎉🎉🎉

I'm unable to build Release mode with .NET Native enabled locally. Something in this PR is causing ILT0005 to happen consistently, every time. 🙁 

But now we have a way to repro this issue locally! It's a small win - I'll liaise with Tom about getting a fix.

michael-hawker:
Marking as draft so we don't accidentally merge. We're pretty much good to go, but this is now reproing the .NET Native issue locally so using it to help accelerate our resolution towards that issue.

PR #355, opened 2023-01-12
[Gallery] UI tweaks

niels9001:
This PR optimizes the sample renderer layout. It increases the default MaxWidth to 600, introduces a ScrollViewer so that samples can be scrolled and optimizes the options layout so it auto collapsed whenever no sample options are defined.

With options
image

No settings defined:
image


michael-hawker (approved):
Looks good, as you mentioned we might have to go through and clean-up some of the samples. (The ContentSizer shelf one is affected I think, at least on WASM).

We should ensure the template is in good standing too.

PR #356, opened 2023-01-13
Sample improvements

niels9001:
Cleaning up the existing samples (e.g. improving colors, removing unnecessary scrollviewers).

Project template looked good, no changes needed there?


michael-hawker:
@niels9001 one thing with something like the CanvasLayout sample is that it also needs Horizontal scrolling...

image

So how should we handle this case?

niels9001:
> @niels9001 one thing with something like the CanvasLayout sample is that it also needs Horizontal scrolling...
> 
> image
> 
> So how should we handle this case?

Oh, I'd be expecting the horizontal scrollbar to show up here (since it works for thr vertical by default as well)? I will have a look!

michael-hawker (approved):
Tested UWP/WASDK/WASM locally, all looks good still.

PR #357, opened 2023-01-17
🧪 [Experiment] Segmented control

niels9001:
Segmented


michael-hawker:
image

Looks like Label is missing on WASM, don't forget to update/clean-up the descriptions and things.

michael-hawker:
Didn't see anything too crazy as it's just a `ContentPresenter`, eh? 🤔

@Arlodotexe now that 4.7 is out, we should probably rev our Uno packages and see if that fixes anything, eh?

michael-hawker:
@niels9001 your PR has been updated and rebased on main and using the new component format! Think everything is working as it should I tested things on UWP/WinAppSDK/WASM. Looks good! (Though WASM wasn't respecting alignment, we'll need to update Uno and check if still an issue.)

Was there anything else to do here or should we merge it? ☺

components/CanvasLayout/samples/Dependencies.props
```diff
@@ -11,21 +11,21 @@
 <Project>
     <!-- WinUI 2 / UWP -->
     <ItemGroup Condition="'$(IsUwp)' == 'true'">
```
michael-hawker, line 13: Ah, oops, this is what happened the rebase changed the wrong file for some weird reason... will see if I can fix...

michael-hawker (approved):
Control looks like a great start, super exciting!

Don't forget to add discussion/issue to the PR description.

components/SegmentedControl/samples/SegmentedControl.md
```diff
@@ -0,0 +1,37 @@
...
+title: SegmentedControl
+author: niels9001
+description: A common UI control to configure a view or setting. 
+keywords: SegmentedControl, Control, Layout
+dev_langs:
+  - csharp
+category: Controls
+subcategory: Layout
+discussion-id: 314
+issue-id: 0
```
michael-hawker, line 11: Don't forget to create a tracking issue

niels9001:
> Control looks like a great start, super exciting!
> 
> Don't forget to add discussion/issue to the PR description.

Only thing I couldn't get to work is setting the SelectedIndex.. (IIRC it works when an ItemsSource is used.. but not when items are hardcoded in XAML).

Setting IsSelected on an item works.. but the SelectedIndex on the Segment control itself does not 😮

michael-hawker:
Hmm, I can take a look later. For now since that wouldn't effect API and we have a valid POC, let's merge it and work forward! That's the whole point of labs! 😀 🎉

PR #367, opened 2023-02-21
[SettingsControls] Using ItemsRepeater

niels9001:


labs/SettingsControls/samples/SettingsControls.Samples/SettingsExpander.md
```diff
@@ -24,3 +24,5 @@ You can easily override certain properties to create custom experiences. For ins
 `SettingsExpander` is also an `ItemsControl`, so its items can be driven by a collection and the `ItemsSource` property. You can use the `ItemTemplate` to define how your data object is represented as a `SettingsCard`, as shown below. The `ItemsHeader` and `ItemsFooter` property can be used to host custom content at the start or end of the items list.
 
 > [!SAMPLE SettingsExpanderItemsSourceSample]
+
+NOTE: Due to an bug, related to the `ItemsRepeater` used in `SettingsExpander`, there might be visual glitches whenever the `SettingsExpander` gets expanded and a `MaxWidth` is set on the parent `StackPanel`. As a workaround, the `StackPanel` (that has the `MaxWidth` set) can be wrapped in a `Grid` to overcome this issue. See the `SettingsPageExample` for snippet.
```
niels9001, line 28: @michael-hawker This probably needs some better rephrasing :-)

michael-hawker:
```suggestion
NOTE: Due to [a bug](https://github.com/microsoft/microsoft-ui-xaml/issues/3842) related to the `ItemsRepeater` used in `SettingsExpander`, there might be visual glitches whenever the `SettingsExpander` expands and a `MaxWidth` is set on a parent `StackPanel`. As a workaround, the `StackPanel` (that has the `MaxWidth` set) can be wrapped in a `Grid` to overcome this issue. See the `SettingsPageExample` for snippet.
```

Note we probably need to adopt the special `> [!Note]` or whatever syntax for docs or look at the new GitHub format for that and set that up to display properly in the app like we do for the current sample app...

michael-hawker (approved):
Weird it didn't credit commit https://github.com/CommunityToolkit/Labs-Windows/pull/367/commits/e53f23d8803fa43a0001c12a78a734006cd8bd04 to me from what I guess was a cherry-pick? Anyway, minor nit.

We also should remove the StyledContentPresenter, but can follow-up with that later, as will just merge this since it built to unblock others for now.

PR #370, opened 2023-02-22
[SettingsControls] Bugfixes

niels9001:
- Fix the FocusVisual on the RightAlignedToggleSwitchStyle
- Fix a visual glitch where the Header and Icon would move whenever the Content was empty


michael-hawker:
@niels9001 looks like this branch isn't based correctly? @Arlodotexe I wonder if the hiccup with the `main` branch in that other PR has caused changes that everyone needs to repull `main`?

niels9001:
Closing this in favor of #373

PR #369, opened 2023-02-21
Organize component folders

Arlodotexe:
Prerequisite PR #360 

This PR introduces the following changes:
- [Renamed ./labs to ./components](https://github.com/CommunityToolkit/Labs-Windows/commit/cb6c66b23607a0aa86b9d69afe90cd767346164c)
- [Moved component Sample and Test files up one directory](https://github.com/CommunityToolkit/Labs-Windows/commit/76d2d7b80e21eb3a0fd604d7c1a8384b96248625)
- Closes #332

PR #360, opened 2023-02-01
Consolidated ProjectHeads and supporting infrastructure

Arlodotexe:
Closes #358

This PR makes several improvements:
- All prop files and scripts related to project heads have been moved to `./common/ProjectHeads/`
- All prop files and scripts related to multitargeting have been moved to `./common/MultiTarget/`
- The all-component heads in `./Platforms` have been moved to `./common/ProjectHeads/AllComponents/`
- The project heads for individual components are now generated from `./common/ProjectHeads/SingleComponent/`.
- The script `./common/ProjectHeads/GenerateSingleSampleHeads.ps1` has been added to copy the single-component heads into a component folder and update the names to reflect the component.
- All non-essential MSBuild declarations in Sample project .csproj files have been moved to `./common/ToolkitComponent.SampleProject.props`.
- All non-essential MSBuild declarations in Source project .csproj files have been moved to `./common/ToolkitComponent.SourceProject.props`.
- The macOS, iOS, GTK, Android and WPF heads have been removed. These used the old Xamarin-based Uno template, and will soon be replaced with the newest Uno template in #256.

PR #372, opened 2023-02-24
Fixed custom csproj file names

Arlodotexe:
This PR is a hotfix for #360 / #369. Rive changed their csproj name, and we had places where we assumed the name of the source project. The assumed name has been replaced with wildcards.

- [Fixed an issue where using a custom file name for the component source csproj prevented the solution from generating](https://github.com/CommunityToolkit/Labs-Windows/commit/ac7520a260894cd813fb36cce1c0e606d876f55b)


michael-hawker (approved):

michael-hawker:
Thanks @Arlodotexe nice find! I like this fix as it generalizes things which should make this flow forward for when we start having a bunch of different names for toolkit packages in outside-labs usages of this stuff!

Issue #362, opened 2023-02-06
Multiple ToolkitBoolOptions causes crash


niels9001:
### Describe the bug

There seems to be an issue that whenever defining two ToolkitBoolOptions the source generator throws an error. Seems that whenever the second bool is bind to it crashes. Probably due to changes made in:  #336

```xml
<labs:SettingsExpander IsExpanded="{x:Bind IsCardEnabled, Mode=OneWay}"/>
```

This works:
```csharp
[ToolkitSampleBoolOption("IsCardEnabled", true, Title = "Is Enabled")]
[ToolkitSampleBoolOption("IsCardExpanded", false, Title = "Is Expanded")]
```

This does not (see change in order): 
```csharp
[ToolkitSampleBoolOption("IsCardExpanded", false, Title = "Is Expanded")]
[ToolkitSampleBoolOption("IsCardEnabled", true, Title = "Is Enabled")]
```

Tested with two ToolkitTextOptions and that seemed to work fine.

`System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
   at SettingsControlsExperiment.Samples.SettingsExpanderSample.get_IsCardEnabled() in C:\Repos\Labs-Windows\labs\SettingsControls\samples\SettingsControls.Samples\CommunityToolkit.Labs.Core.SourceGenerators\CommunityToolkit.Labs.Core.SourceGenerators.ToolkitSampleOptionGenerator\SettingsControlsExperiment.Samples.SettingsExpanderSample.Property.IsCard`

@Arlodotexe FYI

### Steps to reproduce

```text
- Open the SettingsControls experiment on main
- Run samples, navigate to "SettingsExpanderSample"
- Crash
```

### Expected behavior

Not crash

### Code Platform

- [X] UWP
- [ ] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [X] Windows 11 21H2 (Build 22000)
[...]

PR #373, opened 2023-02-24
[SettingsControls] Bugfixes and improvements

niels9001:

**1. Fix for focus visual bug for RightAlignedToggleSwitch**
Before:
image

After:
image

----

**2. Changing UniformGridLayout to StackLayout**
This resolves issues where the content is not always rendered correctly. We were using UniformGridLayout to resolve the expanding bug, but in the end that wasn't working either. StackLayout would be the preferred way to do this anyways.

----

**3. Updating the SettingsExpander content MinHeight**
By default, the Expander.Content MinHeight is enherited from the Header. This resulted, in some cases, that a gap was visible when a SettingsCard would have a lower MinHeight. It's now set to 16 and is overrideable.(`SettingsExpanderContentMinHeight`)

----

**4. Fixing a visual glitch whenever Content is null**
If the content is null, the contentPresenter was still visible. There's a 8px TopMargin used to show a gap.. but that doesn't make sense when Content is null.

Before:
SettingsCardJumpBefore

After:
SettingsCardJumpAfter

**5. Other fixes**
- Removing unused StyledContentPresenter
- Reordered some visual states
- Removing the multiple BoolOptions to avoid the SettingsExpander sample of crashing, until https://github.com/CommunityToolkit/Labs-Windows/issues/362 gets resolved.


components/SettingsControls/src/Helpers/IsNullOrEmptyStateTrigger.cs
```diff
@@ -0,0 +1,141 @@
...
+using Microsoft.Toolkit.Uwp.Helpers;
+#endif
+
+
+namespace CommunityToolkit.Labs.WinUI;
+
+/// <summary>
+/// Enables a state if an Object is <c>null</c> or a String/IEnumerable is empty
+/// </summary>
+public class IsNullOrEmptyStateTrigger : StateTriggerBase
```
Arlodotexe, line 20: We should consider making this internal, to avoid conflicting with helpers in other components down the road (including the inbox `IsNullOrEmptyStateTrigger` that'll get ported soon)

Arlodotexe (approved):
A suggestion, looks good otherwise.

components/SettingsControls/src/Helpers/IsNullOrEmptyStateTrigger.cs
michael-hawker: Can it be referenced from XAML still if `internal`? I thoughts that why we weren't doing it. I mean ideally, we just get these ported over in the next week or two and then just start referencing preview packages for them in labs components?

components/SettingsControls/samples/SettingsExpanderSample.xaml.cs
```diff
@@ -4,7 +4,6 @@
 
 namespace SettingsControlsExperiment.Samples;
 
-[ToolkitSampleBoolOption("IsCardEnabled", true, Title = "Is Enabled")]
```
michael-hawker, line 7: Maybe just comment this out and leave a comment with a link to the issue so we remember to add it back when that issue is fixed?

components/SettingsControls/src/Helpers/IsNullOrEmptyStateTrigger.cs
```diff
@@ -0,0 +1,141 @@
...
+
+    /// <summary>
+    /// Identifies the <see cref="Value"/> DependencyProperty
+    /// </summary>
+    public static readonly DependencyProperty ValueProperty =
+        DependencyProperty.Register(nameof(Value), typeof(object), typeof(IsNullOrEmptyStateTrigger), new PropertyMetadata(null, OnValuePropertyChanged));
+
+    public IsNullOrEmptyStateTrigger()
+    {
+        UpdateTrigger();
```
michael-hawker, line 39: Out of curiosity (since we don't have the diff from the original), was this the only difference to fix the trigger?

components/SettingsControls/src/SettingsExpander/SettingsExpander.xaml
```diff
@@ -263,7 +262,7 @@
                         <Border x:Name="ExpanderContentClip"
                                 Grid.Row="1">
                             <Border x:Name="ExpanderContent"
-                                    MinHeight="{TemplateBinding MinHeight}"
+                                    MinHeight="{ThemeResource SettingsExpanderContentMinHeight}"
```
michael-hawker, line 265: Is this the only place `MinHeight` was bound? It's still in the style for the template. If it was overloaded, this is fine I think. Otherwise, we should repurpose it to represent this and keep the template binding, but change the value in the Style to this resource instead of `{StaticResource ExpanderMinHeight}` on L124 right?

michael-hawker:
Couple of small comments/questions, otherwise should be good. After get those answers, can either make changes or merge as needed.

components/SettingsControls/src/Helpers/IsNullOrEmptyStateTrigger.cs
niels9001: Yep! Tried adding the constructor first and add UpdateTrigger), but then saw that it defaulted to 'true' (bool). Changed this to 'null'and that did the trick.

As discussed, testing these kinds of uses cases would be good when we bring the triggers over :)

components/SettingsControls/src/SettingsExpander/SettingsExpander.xaml
niels9001: Yeah, it's been used for both the toggle button part (= header) and the expandable part (= content). I guess that's the reason why this bug appeared (as a SettingsCard in the content normally has a lower minheight because it has less padding).

components/SettingsControls/src/Helpers/IsNullOrEmptyStateTrigger.cs
niels9001: Yeah the idea is to remove these custom triggers until we have ported them over and made the fixes there.

PR #374, opened 2023-02-24
[SegmentedControl] Adding AutoSelection and high contrast improvements

niels9001:
See #392

- Adds the AutoSelection property (by default the first item will be selected - this can be disabled if needed)
AutoSelection
 

- High contrast improvements
SegmentedControlHCAfter


components/SegmentedControl/src/Segmented/Segmented.Properties.cs
```diff
@@ -0,0 +1,28 @@
...
+    public static readonly DependencyProperty AutoSelectionProperty = DependencyProperty.Register(
+        nameof(AutoSelection),
+        typeof(bool),
+        typeof(Segmented),
+        new PropertyMetadata(defaultValue: true));
+
+    /// <summary>
+    /// Gets or sets the icon.
+    /// </summary>
+    public bool AutoSelection
```
niels9001, line 23: @michael-hawker Do you have a better name for this :)?

components/SegmentedControl/src/Segmented/Segmented.Properties.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// The backing <see cref="DependencyProperty"/> for the <see cref="AutoSelection"/> property.
+    /// </summary>
+    public static readonly DependencyProperty AutoSelectionProperty = DependencyProperty.Register(
+        nameof(AutoSelection),
+        typeof(bool),
+        typeof(Segmented),
+        new PropertyMetadata(defaultValue: true));
+
+    /// <summary>
+    /// Gets or sets the icon.
```
Arlodotexe, line 21:
```suggestion
    /// Gets or sets a value that indicates if the first item should be auto-selected on load.
```

components/SegmentedControl/src/Segmented/Segmented.Properties.cs
michael-hawker: Do we need this property? Since we're inheriting from `ListViewBase` we already have `SelectedItem`/`SelectedIndex`, so if they're not set we don't select anything by default. (i.e. remove the logic around L102 in the segemented.cs)

Then we just fix the issue with those not working for whatever reason instead? Chances can be that someone's going to use those anyway to rehydrate a state so then it'd conflict with having this sort of property as well?

Thoughts?

michael-hawker (changes requested):
Ugh, silly GitHub requires a comment here if you do things out of order... 😋

michael-hawker:
@niels9001 FYI, I noticed you had some usages of `as` with pragmas to ignore warnings of the null deref. Easier pattern is to use the `is` typing like so:

```cs
        if (element is SegmentedItem segmentedItem)
        {
            segmentedItem.Loaded += SegmentedItem_Loaded;
        }
```

michael-hawker:
FYI @niels9001 I noticed at different DPI (150?) the padding is off:

image

image

image

michael-hawker:
Same C# `is` cleanup can be done here:
```cs
            // Only do stuff if the index is actually changing
            if (index != previousIndex)
            {
                if (ContainerFromIndex(index) is SegmentedItem newItem)
                {
                    newItem.Focus(FocusState.Keyboard);
                    retVal = true;
                }                
            }
        }

        return retVal;
    }

    private SegmentedItem? GetCurrentContainerItem()
    {
        if (ControlHelpers.IsXamlRootAvailable && XamlRoot != null)
        {
            return FocusManager.GetFocusedElement(XamlRoot) as SegmentedItem;
        }
        else
        {
            return FocusManager.GetFocusedElement() as SegmentedItem;
        }
    }
```

michael-hawker:
@niels9001 was looking into the selection issue with `SelectedIndex`. I think the issue is that `SelectedIndex`/Item is read first before the `Items` collection has loaded, so there's a timing issue. (Also, through XAML the items load one-by-one I think.)

Was investigating if we could simplify the logic here than trying to check everytime an item is loaded if we've set the index. Will think about it more Monday.

components/SegmentedControl/src/Segmented/Segmented.cs
```diff
@@ -40,18 +26,18 @@ protected override bool IsItemItsOwnContainerOverride(object item)
     protected override void OnApplyTemplate()
     {
         base.OnApplyTemplate();
+        SelectedIndex = _internalSelectedIndex;
```
michael-hawker, line 29: Thinking we should probably only do this once? If the template is re-applied, then the `SelectedIndex` may have changed elsewhere after the initial load? Maybe it's something we just test/keep an eye on later?

michael-hawker (approved):
Tested on UWP and WinAppSDK, looks great.

michael-hawker:
@Arlodotexe feel free to take a quick poke if you'd like and/or just merge.

components/SegmentedControl/src/Segmented/Segmented.cs
niels9001: Good catch - added a hasloaded check, that should fix this?

Arlodotexe (approved):

PR #377, opened 2023-02-24
Workaround for Uno-exposed IDisposable errors in release mode

Arlodotexe:
This PR is a hotfix for all components that may use a Storyboard. Our CI treats all warnings as errors in release mode, so this has been suppressed until Uno can fix it.

From [here](https://github.com/CommunityToolkit/Labs-Windows/pull/275#issuecomment-1331113635):
> Some objects are marked as disposable for internal constraints, but it's not something we can change right away as it may be a breaking change (we'll take note to add it as such for our next major). Having to add `IDisposable` here is likely a result of some analyzer that can be disabled for Uno targets specifically.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -432,3 +432,6 @@ csharp_style_prefer_extended_property_pattern = true:suggestion
 
 # Require file header
 dotnet_diagnostic.IDE0073.severity = warning
+
+# Uno platform exposes IDisposable on Storyboard publicly when it should be internal. Ignore this.
+dotnet_code_quality.CA1001.excluded_type_names_with_derived_types = T:Windows.UI.Xaml.Media.Animation.Storyboard


michael-hawker:
@Arlodotexe isn't their an open Uno issue on this too, can we find that and link the url to the comment in the config too?

I didn't see it linked on our board, so if there is one we should add it there for tracking too: https://github.com/orgs/CommunityToolkit/projects/3/views/6

Arlodotexe:
> @Arlodotexe isn't their an open Uno issue on this too, can we find that and link the url to the comment in the config too?
> 
> I didn't see it linked on our board, so if there is one we should add it there for tracking too: https://github.com/orgs/CommunityToolkit/projects/3/views/6

Going back to the original comments from @jeromelaban and looking through open Uno issues, I wasn't able to find an existing issue for this.

michael-hawker:
@Arlodotexe, want to open one so we can link it to all the places and track then? I know @jeromelaban mentioned it'd take a version to fix or something as it'd be a breaking change. Not sure if he meant a 4.8 or a 5.0 sort of thing though.

jeromelaban:
@michael-hawker `IDisposable` on types exposed on iOS, Android and Catalyst will continue to do so (it's coming from .NET bindings to native UIKit/Java types) and this not something we'll change any time soon. This is a characteristic of Uno's Visual Tree inheriting directly from native types for performance considerations.

Arlodotexe:
Understandable, we'll accommodate then and continue suppressing the warning. @michael-hawker want to sign off on this?

jeromelaban:
@Arlodotexe It could be something possible, though it would be a general exclusion of that warning regardless of the location. Unless I'm mistaken, this happens at the reference site and I don't think analyzers allow this kind of flexibility.

michael-hawker:
Random idea I had based on this discussion and the Marquee PR. We could use the Uno props file to only ignore the warning for Uno builds, like this:

https://github.com/CommunityToolkit/Labs-Windows/blob/10640a375501b4b38ce892c5e9990213d76812f6/common/MultiTarget/Uno.props#L23-L26

If there's still a disposable issue it should be flagged from UWP/WinAppSDK bits still, but at least not flag it for other places based on this type of issue, eh?

Arlodotexe:
> Random idea I had based on this discussion and the Marquee PR. We could use the Uno props file to only ignore the warning for Uno builds, like this:
> 
> https://github.com/CommunityToolkit/Labs-Windows/blob/10640a375501b4b38ce892c5e9990213d76812f6/common/MultiTarget/Uno.props#L23-L26
> 
> If there's still a disposable issue it should be flagged from UWP/WinAppSDK bits still, but at least not flag it for other places based on this type of issue, eh?

I like this idea. If we globally-suppress _only_ on Uno, this error will show on the UWP / WinAppSDK heads instead, and users won't have to manually suppress anything for their control like MarqueeText did. 

The only problem with this I can foresee is that a component which _exclusively_ targets Uno won't see this error and may end up with a memory leak. All of our components target UWP or WinAppSDK, so this isn't a problem for now.

We'll move forward with this as a workaround.

PR #379, opened 2023-03-01
[Experiment] Shimmer

niels9001:
Adding the Shimmer control from the Store. See: https://github.com/CommunityToolkit/Labs-Windows/discussions/381

Shimmerrr


components/Shimmer/src/Shimmer/Shimmer.cs
```diff
@@ -0,0 +1,237 @@
...
+    {
+        ActualThemeChanged -= OnActualThemeChanged;
+        StopAnimation();
+
+        if (_initialized && _shape != null)
+        {
+#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
+            ElementCompositionPreview.SetElementChildVisual(_shape, null);
+#pragma warning restore CS8625 // Cannot convert null literal to non-nullable reference type.
+            _disposableVisualResources?.Dispose();
```
niels9001, line 94: Note to self: dispose elements here manually

michael-hawker:
@Arlodotexe found out what happens if you try and run WASM head, it just says that the underlying project isn't compatible with the TFM... But since we don't build those the same way in the CI, it works fine here. So just something we can improve later for the `OpenSolution` solution...

components/Shimmer/src/Shimmer/Shimmer.Properties.cs
```diff
@@ -0,0 +1,46 @@
...
+                {
+                    self.TryStartAnimation();
+                }
+                else
+                {
+                    self.StopAnimation();
+                }
+            }));
+
+    public double AnimationDurationInMilliseconds
```
michael-hawker, line 34: Should we just call this `AnimationDuration` (or just even `Duration` as it's 'shimmer.Duration') and make it a `TimeSpan` like we did with the `Duration` property with the Animation APIs in the Toolkit?

e.g. https://learn.microsoft.com/dotnet/api/microsoft.toolkit.uwp.ui.animations.animationscope.duration (doesn't have to be nullable as we provide a default value) or https://learn.microsoft.com/windows/winui/api/microsoft.ui.xaml.controls.ianimatedvisual.duration

components/Shimmer/src/Shimmer/Shimmer.cs
```diff
@@ -0,0 +1,234 @@
+
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Numerics;
+
+
```
michael-hawker, line 8:
```suggestion

```

extra newline

components/Shimmer/src/Shimmer/Shimmer.cs
```diff
@@ -0,0 +1,234 @@
...
+using Windows.UI.Composition;
+using Windows.UI.Xaml.Hosting;
+using Windows.UI.Xaml.Shapes;
+using Microsoft.Toolkit.Uwp.UI;
+using Microsoft.Toolkit.Uwp.UI.Animations.Expressions;
+using Microsoft.Toolkit.Uwp.UI.Animations;
+#endif
+
+
+namespace CommunityToolkit.Labs.WinUI;
```
michael-hawker, line 28: Should we try the new `CommunityToolkit.WinUI.Controls` namespace here yet?

components/Shimmer/src/Shimmer/Shimmer.cs
```diff
@@ -0,0 +1,234 @@
...
+
+/// <summary>
+/// A generic shimmer control that can be used to construct a beautiful loading effect.
+/// </summary>
+[TemplatePart(Name = PART_Shape, Type = typeof(Rectangle))]
+#pragma warning disable CA1001 // Types that own disposable fields should be disposable
```
michael-hawker, line 34: Should be resolved once we fix #377 eh?

components/Shimmer/src/Shimmer/Shimmer.cs
```diff
@@ -0,0 +1,234 @@
...
+    private CompositionColorGradientStop _gradientStop4;
+    private CompositionRoundedRectangleGeometry _rectangleGeometry;
+    private ShapeVisual _shapeVisual;
+    private CompositionLinearGradientBrush _shimmerMaskGradient;
+    private Border? _shape;
+
+    private bool _initialized;
+    private bool _animationStarted;
+
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
```
michael-hawker, line 56: We should probably make everything probably nullable?

components/Shimmer/src/Shimmer/Shimmer.Properties.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        get => (double)GetValue(AnimationDurationInMillisecondsProperty);
+        set => SetValue(AnimationDurationInMillisecondsProperty, value);
+    }
+
+    public bool IsAnimating
```
michael-hawker, line 40: Should we use `IsActive` to mimic `ProgressRing`?

https://learn.microsoft.com/windows/winui/api/microsoft.ui.xaml.controls.progressring.isactive

components/Shimmer/src/Shimmer/Shimmer.xaml
```diff
@@ -0,0 +1,38 @@
...
+        <Setter Property="CornerRadius" Value="4" />
+        <Setter Property="MinWidth" Value="40" />
+        <Setter Property="MinHeight" Value="8" />
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="labs:Shimmer">
+                    <Border x:Name="Shape"
+                            Background="{TemplateBinding Background}"
+                            CornerRadius="{TemplateBinding CornerRadius}" />
+                </ControlTemplate>
```
michael-hawker, line 34: So little XAML. As another follow-up for the tracking issue, maybe should investigate if we can go the `ContentPresenter` route like we did with Constrained Box and do some funky stuff like that... Trying to think what the most lightweight element we can inherit from would be since `Border` is sealed... As we could take a couple of approaches, but would be nice to remove the XAML template as then would be super light-weight.

ContentPresenter would lend itself to just wrapping something else, though I know our attached shadow like effect route could be interesting as well. Probably different use cases depending on what and how things are being loaded. We'll definitely have to play around more with these ideas in the future.

michael-hawker (changes requested):
Looks good, tested on UWP and WindowsAppSDK without issue (so cool to see UWP code just working over there via Labs).

Have some suggestions for the property names and future investigations we should track.

components/Shimmer/samples/Shimmer.md
```diff
@@ -2,7 +2,7 @@
 title: Shimmer
 author: JustinXinLiu
```
michael-hawker, line 3: Saw this was the commit you got the --author to work, though we should try and get it on the init commit where ever the bulk of code was imported.

components/Shimmer/src/Shimmer/Shimmer.Properties.cs
niels9001: Agree!

Issue #190, opened 2022-07-07
Can't run in VS Code Locally


michael-hawker:
Looks like the revamp in #91 to make everything work in codespaces broke things from working in VS Code locally due to the use of `pwsh` in the `tasks.json` pre-launch step:

```
> Executing task: pwsh ./GenerateAllSolution.ps1 <

pwsh : The term 'pwsh' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was 
included, verify that the path is correct and try again.
At line:1 char:1
+ pwsh ./GenerateAllSolution.ps1
+ ~~~~
    + CategoryInfo          : ObjectNotFound: (pwsh:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
 
The terminal process "C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe -Command pwsh ./GenerateAllSolution.ps1" terminated with exit code: 1.

Terminal will be reused by tasks, press any key to close it.
```

We need to figure out a way to make this work for both local and Codespace development on Windows/Linux, there must be something to handle this built into VS Code itself, right?

1. Clone main
2. Open up VS Code in root
3. Run and Debug 'All samples app'
4. See error.

On Win 10 latest.

Arlodotexe:
This should help: https://stackoverflow.com/a/64911186/5953220

PR #380, opened 2023-03-02
Fixed building in codespaces

Arlodotexe:
In #360 and #369, we organized and consolidated our projects heads, and renamed the folder which components are contained in.

While these were tested to work locally, Codespaces was not reconfigured to use the new paths.

This PR
- Fixes sample discovery in codespaces.
- Fixes head generation in codespaces.
- Fixes building and deploying in codespaces.
- Fixed an issue where a missing `Path` environment variable would cause the build to fail (e.g., on linux)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -14,13 +14,13 @@
         "/p:UnoSourceGeneratorUseGenerationHost=true",
         "/p:UnoSourceGeneratorUseGenerationController=false",
         "/p:UnoRemoteControlPort=443",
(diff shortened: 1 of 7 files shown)


michael-hawker (approved):
Tested in codespaces, works great! 🎉🎉🎉

michael-hawker:
@Arlodotexe does this fix #190 as well now then too?

Arlodotexe:
> @Arlodotexe does this fix #190 as well now then too?

That's a bit of a different issue.

PR #382, opened 2023-03-03
[SettingsExpander] Adding Expanded/Collapsed events

niels9001:
To solve: https://github.com/CommunityToolkit/Labs-Windows/discussions/364

```xml
<SetingsExpander Expanded="OnExpanded" Collapsed="OnCollapsed" .... />
```


michael-hawker (approved):
Suppose this is simple enough to just include. Was thinking we could just show a sample or something of this for the request. But this is pretty minimal to just have and makes sense. `Expander` in general doesn't have the largest surface area of unique APIs.

Will be interesting to see if I can get us the `Expanding` event later with the work I'm looking at for complementing the platform expander [in this branch here](https://github.com/CommunityToolkit/Labs-Windows/tree/llama/expander-ex/labs/Expander).

Issue #390, opened 2023-03-08
🧪 [Experiment] Shimmer


niels9001:
### Approved from Discussion

https://github.com/CommunityToolkit/Labs-Windows/discussions/381

### Problem Statement

A way to indicate to the user that certain parts of the screen are loading.

### Overview

This experiment adds the following components:
 - **Shimmer**: a loading indication control

### Using

You can try it out via the NuGet Packages here:

UWP: https://dev.azure.com/dotnet/CommunityToolkit/_artifacts/feed/CommunityToolkit-Labs/NuGet/CommunityToolkit.Labs.Uwp.CanvasView
WinUI 3: https://dev.azure.com/dotnet/CommunityToolkit/_artifacts/feed/CommunityToolkit-Labs/NuGet/CommunityToolkit.Labs.WinUI.CanvasView
Read more about [Preview Packages here](https://aka.ms/wct/wiki/previewpackages).

CommunityToolkit members can also try it out with Codespaces.

### Additional info
TO DO:
- [ ] Add sample of using an ItemSelector to swap out a loading template with the loaded content.
- [ ] 

FUTURE:
- It'd be really cool to investigate if we could make this easier to use, though I think that'll have to be an 8.1 thing, and this is basically just the raw component that knows how to fill a space and shimmer, eh?
- Can we skip the XAML resource file and do something with a `ContentPresenter` or make it attachable to an UI element (like `AttachedShadow`?

### Implementation Requirements

- [X] Working Prototype
- [X] Feature Complete
- [X] Documentation
- [X] Samples
- [ ] Unit Tests
- [ ] Community Feedback / Usage Testimonies

### Tested Platforms

- [X] UWP
- [X] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Technical Review

- [ ] Accessibility Audit
- [ ] API/Naming Review
- [ ] Code Quality/Style
- [ ] Dependency Review
- [ ] Design/Style Review
- [ ] Final Approval

### Community Help?

Yes

dpaulino:
I'm using this and it works great. Question tho: how to make a perfectly round shimmer?

JohnnyWestlake:
Suggestion: turn this into a Brush instead, allowing all instances to share the same internal composition animation and brush to optimise performance and resource usage.

Issue #387, opened 2023-03-07
Remove Labs from the default namespace for new components


Arlodotexe:
### Describe the bug

The default templated namespace for all components uses the Labs moniker, which won't work for the main toolkit.

After [reviewing past discussions](https://github.com/CommunityToolkit/WindowsCommunityToolkit/issues/4488#issuecomment-1458638688), @michael-hawker, @niels9001 and @Arlodotexe have settled on:
- `CommunityToolkit.Labs.WinUI.Controls.ComponentName` for the package ID.
- `CommunityToolkit.WinUI.Controls.ComponentName` for the default namespace.

Reasoning behind this:
- Our tooling has a major focus on UI controls (tests, samples, etc)
- Namespaces and package IDs can be customized per component anyway, giving us the flexibility to use something different later (see #386).

### Steps to reproduce

```text
n/a
```

### Expected behavior

n/a

### Code Platform

- [ ] UWP
- [ ] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [ ] Windows 11 21H2 (Build 22000)
- [ ] Other (specify)

### App minimum and target SDK version

- [ ] Windows 10, version 1809 (Build 17763)
- [ ] Windows 10, version 1903 (Build 18362)
- [ ] Windows 10, version 1909 (Build 18363)
- [ ] Windows 10, version 2004 (Build 19041)
- [ ] Other (specify)

### Help us help you

Yes, but only if others can assist.

Issue #331, opened 2022-12-08
Hard to use DispatcherQueue/Timer in Labs project


michael-hawker:
### Describe the bug

These are in two different namespaces `Windows.System` for UWP and Microsoft.UI.Dispatching` for WinAppSDK.

We currently don't include these in our global using in https://github.com/CommunityToolkit/Labs-Windows/blob/main/common/GlobalUsings_WinUI.cs

This means it's not easy to just reference `DispatcherQueue` directly in a Labs experiment, which is probably going to be pretty common.

We should add these namespaces to the global usings and make sure they're easy to use.

I did the same in the Stacked Notifications Behavior here:

https://github.com/CommunityToolkit/Labs-Windows/blob/9045236e77884e6c5c9fccb261e390001c93d4ec/labs/StackedNotificationsBehavior/src/StackedNotificationsBehavior.cs#L5-L13

Which Rive needed as well.

I'm not as sure why I couldn't just include the root namespace for the DispatcherQueue or if that was just an oversight here when initially diagnosing the issue.

### Steps to reproduce

```text
Try and use a DispatcherQueue in an exeperiment
```

### Expected behavior

Shouldn't need to include extra namespaces

### Code Platform

- [X] UWP
- [X] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [ ] Windows 11 21H2 (Build 22000)
- [ ] Other (specify)

### App minimum and target SDK version

- [ ] Windows 10, version 1809 (Build 17763)
- [ ] Windows 10, version 1903 (Build 18362)
- [ ] Windows 10, version 1909 (Build 18363)
- [ ] Windows 10, version 2004 (Build 19041)
- [ ] Other (specify)

### Help us help you

Yes, but only if others can assist.

michael-hawker:
Thinking the alias of just `DispatcherQueue` should work? As then anything referencing that type should just be able to be still called `DispatcherQueue` as the type...

PR #391, opened 2023-03-08
[Experiment] Shimmer

niels9001:
Adding the Shimmer control from the Store. See: https://github.com/CommunityToolkit/Labs-Windows/discussions/381 and #390

Shimmerrr


components/Shimmer/src/Shimmer/Shimmer.Properties.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace CommunityToolkit.Labs.WinUI;
+public partial class Shimmer : Control
+{
+    public static readonly DependencyProperty DurationProperty = DependencyProperty.Register(
+        nameof(Duration), typeof(TimeSpan), typeof(Shimmer), new PropertyMetadata(TimeSpan.FromMilliseconds(1600), (s, e) =>
+        {
+            var self = (Shimmer)s;
+            if (self.IsActive)
+            {
+                self.TryStartAnimation();
```
michael-hawker, line 14: Do we need to ensure we stop the animation first? (as doesn't this method no-op if it's already running?)

Could we just update the `_gradient*PointAnimation.Duration` properties, since we have a reference now, would that work?

components/Shimmer/src/Shimmer/Shimmer.cs
```diff
@@ -0,0 +1,228 @@
...
+    private ShapeVisual _shapeVisual;
+    private CompositionLinearGradientBrush _shimmerMaskGradient;
+    private Border? _shape;
+
+    private bool _initialized;
+    private bool _animationStarted;
+
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+    public Shimmer()
+#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
```
michael-hawker, line 54: Bringing over comment from other PR, that we should investigate clearing up this warning by using nullables.

michael-hawker (changes requested):

components/Shimmer/src/Shimmer/Shimmer.cs
niels9001: I made those nullable and used https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving as I wanted to avoid having a bunch of null checks.. but not sure if that's the right approach? Let me know!

components/Shimmer/src/Shimmer/Shimmer.Properties.cs
niels9001: Stopping the animation now first - wrt duration, you want to set those to 0 instead?

common/GlobalUsings_WinUI.cs
```diff
@@ -7,7 +7,7 @@
 
 global using System.Runtime.InteropServices.WindowsRuntime;
 
-global using CommunityToolkit.Labs.WinUI;
```
niels9001, line 10: @Arlodotexe @michael-hawker Should I be touching this? Doesn't compile if not..

components/Shimmer/src/Shimmer/Shimmer.Properties.cs
michael-hawker: Sorry should have made a separate comment. Let me do that...

common/GlobalUsings_WinUI.cs
michael-hawker: @Arlodotexe we should remove this as part of #394

components/Shimmer/src/Shimmer/Shimmer.Properties.cs
michael-hawker: Well, I think there's two approaches here when the `Duration` property changes, we either just stop/restart the animation to update everything.

Otherwise, I was wondering if we could just modify the duration of in the inner animations:

```cs
_gradientStartPointAnimation.Duration = e.NewValue;
_gradientEndPointAnimation.Duration = e.NewValue;
```

But not sure if we'd have to stop and restart the composition animation anyway?

```cs
        _shimmerMaskGradient!.StopAnimation(nameof(CompositionLinearGradientBrush.StartPoint));
        _shimmerMaskGradient.StopAnimation(nameof(CompositionLinearGradientBrush.EndPoint));
        _shimmerMaskGradient!.StartAnimation(nameof(CompositionLinearGradientBrush.StartPoint), _gradientStartPointAnimation);
        _shimmerMaskGradient.StartAnimation(nameof(CompositionLinearGradientBrush.EndPoint), _gradientEndPointAnimation);
```

Which means it may just be easier/simpler to call Stop and then start. Which also means we could abstract the callback here and re-use as a method for both the `IsActive` property changed and the `Duration` one, maybe that's the simplest?

common/GlobalUsings_WinUI.cs
Arlodotexe: @michael-hawker Sure, we can try to get both #321 and #387 into #394.

Only note - https://github.com/CommunityToolkit/Labs-Windows/issues/321#issuecomment-1341776315
> We should fix/test https://github.com/CommunityToolkit/Labs-Windows/issues/331 alongside this transition.

We'll do #331 separately another time.

michael-hawker: Thanks @Arlodotexe, #331 will probably come up a lot for the Toolkit components. If we can agree on a name for the alias (wonder if just `DispatcherQueue` can work?) then would be an easy add if we move the globals, we can investigate if we add that to the PR.

Arlodotexe: Since we're at a critical point for #398 I'd like to close things off quicker than that. There's only pros to closing these off back to back, if we stay on top of it.

components/Shimmer/src/Shimmer/Shimmer.cs
```diff
@@ -0,0 +1,228 @@
...
+#else
+using Windows.UI.Composition;
+using Windows.UI.Xaml.Hosting;
+using Windows.UI.Xaml.Shapes;
+using Microsoft.Toolkit.Uwp.UI;
+using Microsoft.Toolkit.Uwp.UI.Animations.Expressions;
+using Microsoft.Toolkit.Uwp.UI.Animations;
+#endif
+
+namespace CommunityToolkit.Labs.WinUI;
```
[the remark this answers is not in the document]
niels9001: Reverted for now!

common/GlobalUsings_WinUI.cs
Arlodotexe: #394 will have a required fix for `SegmentedControl` not pushing to Nuget ([see logs](https://github.com/CommunityToolkit/Labs-Windows/actions/runs/4377944010/jobs/7662085789#step:12:1292)). This is blocking, so we'll close off #394 once the ticket is filled and just open another for #321.

components/Shimmer/src/Shimmer/Shimmer.Properties.cs
niels9001: Moved the start/stop to a single method 👍

michael-hawker (approved):
Cool, we should be all set, I think! 🎉🎉🎉

Arlodotexe (approved):
Tested and approved, let's ship it! 🥳

PR #396, opened 2023-03-09
[Gallery] Fix for WASDK titlebar brush

niels9001:
Before:
image

After:
image


michael-hawker (approved):
Tested on Win10 with UWP and WASDK for light and dark mode, looks great!

PR #403, opened 2023-03-13
Move project template from /template/ProjectTemplate to /common/ProjectTemplate

### LINKED ISSUE #397: Move ProjectTemplate (from template) into the common folder
We need to move the `ProjectTemplate` folder from the `template` folder directly into the `common` folder.

The `readme` will need find a new home. Probably should move it within the `ProjectTemplate` folder and either exclude it from being copied, or maybe even have it be part of the new template? We could use a similar approach to my winui2+3 template here: https://github.com/michael-hawker/WinUI2Plus3Template/blob/main/.template.config/template.json#L71-L77

Depends on #387

michael-hawker:
Fixes #397

Should rebase and wait until #356 is merged.

This just moves the `ProjectTemplate` folder into the `common` folder (goes well next to `ProjectHeads` anyway).

Had to update a couple of paths, but not many, hopefully everything should still just work, CI should be a telling factor.

We discussed renaming the `common` folder to `tooling`, think will do that as a follow-on PR?

Readme location moved, but was already excluded in template config, so should be good there.


michael-hawker:
Merged #356 and rebased this PR. Had a clean build before, so should be all set to go. Tested template locally, and we have the CI step for that too, so we should be all good.

@Arlodotexe any feedback or concerns? Think we'll be ready to go with this! 🎉🎉🎉

Arlodotexe (approved):
Fully tested on my local machine. Lgtm!

PR #408, opened 2023-03-15
[Segmented] Fix crash due to infinite width

niels9001:
We didn't handle a potential Infinite width on the AvailableSize.. (which happen in very specific edge cases where this control would be in e.g. a Grid.Column set to Auto. This change makes sure it has a proper fallback.


components/SegmentedControl/src/EqualPanel.cs
```diff
@@ -66,25 +74,6 @@ protected override Size ArrangeOverride(Size finalSize)
         }
         return finalSize;
     }
```
michael-hawker, line 76: @niels9001 Note: for your other issue, you're ignoring `finalSize` in the arrange method, so that's why it's not working in the other scenario.

components/SegmentedControl/src/EqualPanel.cs
```diff
@@ -37,18 +37,26 @@ protected override Size MeasureOverride(Size availableSize)
...
+                child.Measure(availableSize);
+                maxItemWidth = Math.Max(maxItemWidth, child.DesiredSize.Width);
+                maxItemHeight = Math.Max(maxItemHeight, child.DesiredSize.Height);
+            }
 
-            // Equal columns based on the available width
-            if (HorizontalAlignment == HorizontalAlignment.Stretch)
+            // Return equal widths based on the widest item
+            // In very specific edge cases the AvailableWidth might be infinite resulting in a crash.
+            if (HorizontalAlignment != HorizontalAlignment.Stretch || double.IsInfinity(availableSize.Width))
```
michael-hawker, line 49: Future: Thinking long-term, maybe we should create an issue to track adding an `Orientation` property here so it can do equal heights as well? Then this would be VerticalAlignment and height as a flip on Orientation here as well?

I'm trying to remember why `UniformGrid` didn't work again, but that's a lot heavier-weight anyway... so simple panel is simple. 😋

components/SegmentedControl/src/EqualPanel.cs
```diff
@@ -37,18 +37,26 @@ protected override Size MeasureOverride(Size availableSize)
 
         if (Children.Count > 0)
         {
-            SetMaxDimensions(Children, availableSize);
+            foreach (var child in Children)
+            {
+                child.Measure(availableSize);
```
michael-hawker, line 42: We should account for collapsed children here as well, eh?

components/SegmentedControl/src/EqualPanel.cs
```diff
@@ -37,18 +37,26 @@ protected override Size MeasureOverride(Size availableSize)
...
+                maxItemHeight = Math.Max(maxItemHeight, child.DesiredSize.Height);
+            }
 
-            // Equal columns based on the available width
-            if (HorizontalAlignment == HorizontalAlignment.Stretch)
+            // Return equal widths based on the widest item
+            // In very specific edge cases the AvailableWidth might be infinite resulting in a crash.
+            if (HorizontalAlignment != HorizontalAlignment.Stretch || double.IsInfinity(availableSize.Width))
+            {
+                return new Size((maxItemWidth * Children.Count) + (Spacing * (Children.Count - 1)), maxItemHeight);
```
michael-hawker, line 51: Uncollapsed children here too

michael-hawker (commented):
Didn't have a chance to test directly, but added some general comments for some of the issues I know we've encountered (or will encounter).

Been thinking we probably need some panel helper methods (like getting count of uncollapsed children)...

niels9001:
This branch got polluted when rebasing via the GitHub went wrong 😣 - new PR: https://github.com/CommunityToolkit/Labs-Windows/pull/408

PR #409, opened 2023-03-15
Moved tooling folder to submodule

Arlodotexe:
## Overview
This PR removes the contents of `./tooling`, and re-adds it via our new [tooling repo](https://github.com/CommunityToolkit/Tooling-Windows-Submodule).

Users will need to use `git pull --recurse-submodules` instead of just `git pull` to bring the submodule into the local working tree.


.github/workflows/build.yml
```diff
@@ -37,6 +37,8 @@ jobs:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - name: Checkout Repository
         uses: actions/checkout@v3
+        with:
+          submodules: recursive
```
michael-hawker, line 41: That's nice and simple 🙂

.github/workflows/build.yml
michael-hawker, line 1: Remove the `new-experiment` and `project-template` jobs as those run out of the Tooling repo now, we don't need to do that all the time here anymore, just the main component build.

`wasm-linux` **can stay** as that includes the components and is super quick anyway.

.gitmodules
michael-hawker, line 1: Add a comment near the top of the root readme about how to properly check this repo out like you did in the PR comment about the git pull with the recurse parameter.

michael-hawker (changes requested):

.github/workflows/build.yml
Arlodotexe: Good call, will do

michael-hawker: Edited comment, forgot I split that as two jobs in the new place for the template build itself and the new experiment creation, in either case, we're all set here to minimize duplicated effort, yay! 🎉🎉🎉

Thanks for the quick turnaround.

michael-hawker, line 1: Ah, remembered that `labexp` template name is now `ctk-component`. Should be fine here now that we removed the `new-experiment` job, but not sure if we call that out in the readme anywhere as well?

michael-hawker (approved):

michael-hawker:
Updated branch protections to not wait on the `new-experiment` step that doesn't exist.

WinUI 2 build failed due to https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/9

michael-hawker:
Huh, build failed but I couldn't get logs from it... assume same issue? Re-running... 😟

michael-hawker:
Thought we disabled the SG tests here?

```
  Failed PaneOption_GeneratesTitleProperty [718 ms]
  Error Message:
   Test method CommunityToolkit.Tooling.SampleGen.Tests.ToolkitSampleMetadataTests.PaneOption_GeneratesTitleProperty threw exception: 
System.InvalidOperationException: Sequence contains no matching element
  Stack Trace:
      at System.Linq.ThrowHelper.ThrowNoMatchException()
   at CommunityToolkit.Tooling.SampleGen.Tests.ToolkitSampleMetadataTests.VerifyGenerateSources(String assemblyName, String source, IIncrementalGenerator[] generators, Boolean ignoreDiagnostics, ValueTuple`2[] results) in /_/CommunityToolkit.Tooling.SampleGen.Tests/ToolkitSampleMetadataTests.cs:line 533
   at CommunityToolkit.Tooling.SampleGen.Tests.ToolkitSampleMetadataTests.PaneOption_GeneratesTitleProperty() in /_/CommunityToolkit.Tooling.SampleGen.Tests/ToolkitSampleMetadataTests.cs:line 91
```

They're run in the tooling repo now, so we shouldn't have to run these tests here too.

PR #411, opened 2023-03-19
Update ReadMe.md

kmgallahan:
Add additional required steps to run the batch file.


ReadMe.md
```diff
@@ -68,7 +68,17 @@ Open up an issue on the main Toolkit repo using the `Toolkit Labs Transfer` Issu
...
-You can build the main Sample App solution to see all the experiments currently available in this repository by running the `GenerateAllSolution.bat` script in the repo root. Then just open the `CommunityToolkit.AllComponents.sln` solution in Visual Studio. You can run one of the project heads such as `CommunityToolkit.App.WinAppSdk` to run the sample app for that platform.
+You can build the main Sample App solution to see all the experiments currently available in this repository by running the `GenerateAllSolution.bat` script in the repo root. 
+
+> **Note**
+>
+> The script requires that you:
+>
+> 1. Install SlnGen by running `dotnet tool install --global Microsoft.VisualStudio.SlnGen.Tool`
+> 2. Run `dotnet tool restore`
+> 3. Add `MSBuild.exe` to your PATH
```
michael-hawker, line 79: Thanks @kmgallahan for the update here. We should probably move extra requirements to the `Build Requirements` section instead, as it's not limited to the sample app.

Note, we have a .NET tooling config file, so you just need to run `dotnet tool restore` in the root of the repo to get the `slngen` tool and don't need the extra first step.

Also, if you run from the `Developer Command Prompt for VS` then msbuild will already be part of the path, so I'd probably phrase it like

> Run from Developer Command Prompt for VS or include MSBuild location in PATH

kmgallahan: Tweaked the text.

Arlodotexe (approved):

PR #413, opened 2023-04-05
[SettingsControls] Accessibility improvements

niels9001:
This PR solves the following issues:

## 1. Adding `IsActionIconVisible`
The `ActionIcon` can now be collapsed when not needed, similar to `InfoBar`.

## 2. SettingsExpander: incorrect controltype
Was using `Button` - resulting in missing Narrator behavior per Accessibility Insights. Should be `Group` as this control is not a button. (https://github.com/microsoft/PowerToys/issues/24800)

## 3. SettingsCard: incorrect focus hierarchy and visual when clickable
Design and a11y team recommended to set the focus visual on the card, and not on the inner `ContentControl`. The `ContentControl` is removed as it has no use anymore.
The focus visual is now on the card itself (that can be clicked) and is now consistent with `SettingsExpander`. This also solves an a11y issue flagged by Accessibility Insights. (https://github.com/microsoft/PowerToys/issues/24807 / https://github.com/microsoft/PowerToys/issues/24803 / https://github.com/microsoft/PowerToys/issues/24802)

Before:
FocusBefore

After:
FocusAfter

## 4. SettingsCard: PressedState triggered when the control in its content is pressed.
Whenever the focus was on the Content control, and the user would use space/enter to active the control, the pressed state of the entire card would be triggered which is incorrect. On keydown, the type of the focused element is now checked first before setting the PressedState.

Before:
PressedBefore

After:
PressedAfter

### DIFF
--- a/components/SettingsControls/src/Helpers/ControlHelper.cs
+++ b/components/SettingsControls/src/Helpers/ControlHelper.cs
@@ -1,0 +1,9 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace CommunityToolkit.Labs.WinUI;
+internal static partial class ControlHelpers
+{
+    internal static bool IsXamlRootAvailable { get; } = Windows.Foundation.Metadata.ApiInformation.IsPropertyPresent("Windows.UI.Xaml.UIElement", "XamlRoot");
+}
--- a/components/SettingsControls/src/SettingsCard/SettingsCard.Properties.cs
+++ b/components/SettingsControls/src/SettingsCard/SettingsCard.Properties.cs
@@ -59,7 +59,6 @@ public partial class SettingsCard : ButtonBase
         typeof(SettingsCard),
         new PropertyMetadata(defaultValue: false, (d, e) => ((SettingsCard)d).OnIsClickEnabledPropertyChanged((bool)e.OldValue, (bool)e.NewValue)));
 
-
     /// <summary>
     /// The backing <see cref="DependencyProperty"/> for the <see cref="ContentAlignment"/> property.
     /// </summary>
@@ -68,6 +67,15 @@ public partial class SettingsCard : ButtonBase
         typeof(ContentAlignment),
         typeof(SettingsCard),
         new PropertyMetadata(defaultValue: ContentAlignment.Right));
+
+    /// <summary>
+    /// The backing <see cref="DependencyProperty"/> for the <see cref="IsActionIconVisible"/> property.
+    /// </summary>
+    public static readonly DependencyProperty IsActionIconVisibleProperty = DependencyProperty.Register(
+        nameof(IsActionIconVisible),
+        typeof(bool),
+        typeof(SettingsCard),
+        new PropertyMetadata(defaultValue: true, (d, e) => ((SettingsCard)d).OnIsActionIconVisiblePropertyChanged((bool)e.OldValue, (bool)e.NewValue)));
 
     /// <summary>
     /// Gets or sets the Header.
@@ -134,6 +142,15 @@ public partial class SettingsCard : ButtonBase
         set => SetValue(ContentAlignmentProperty, value);
     }
 
+    /// <summary>
+    /// Gets or sets if the ActionIcon is shown.
+    /// </summary>
+    public bool IsActionIconVisible
+    {
+        get => (bool)GetValue(IsActionIconVisibleProperty);
+        set => SetValue(IsActionIconVisibleProperty, value);
+    }
+
     protected virtual void OnIsClickEnabledPropertyChanged(bool oldValue, bool newValue)
     {
         OnIsClickEnabledChanged();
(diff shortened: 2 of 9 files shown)


components/SettingsControls/src/Helpers/ControlHelper.cs
```diff
@@ -0,0 +1,9 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace CommunityToolkit.Labs.WinUI;
+internal static partial class ControlHelpers
+{
+    internal static bool IsXamlRootAvailable { get; } = Windows.Foundation.Metadata.ApiInformation.IsPropertyPresent("Windows.UI.Xaml.UIElement", "XamlRoot");
```
niels9001, line 8: @michael-hawker This helpers comes back in other controls as well (e.g. TokenView). Is this something we could put into a common library?

michael-hawker: It's a pain as we target 1809, but the XamlRoot was only added in 1903... I mean it's a simple shortcut, so don't know if worth a common helper and taking a dependency, unless we do like a Source Generator magic sort of thing... 😋 But that's not simple either.

We could also try and up our min SDK, but then we wouldn't be aligned with the WinAppSDK side, though it supports this property... 21H2 is the next LTSC release that's out now.

But less than 2% of folks are on 1809 or older, most are on 1903 and above, latest report I could find from middle of last year: https://reports.adduplex.com/#/r/2022-06

image

michael-hawker: For now, let's just keep copying this simpler helper as needed.

@Sergio0694 is PolySharp just for .NET runtime stuff or could it or something similar be used to generate this type of thing as well?

Sergio0694: Honestly I'd just bump to 1903 (18362) everywhere 😄

michael-hawker: @Sergio0694 yeah, it's a bit odd as I think WinUI 3 supports down to 1809, so then we'd be mismatched for LTSC for UWP there, that's my only concern really.

components/SettingsControls/src/SettingsCard/SettingsCard.xaml
```diff
@@ -439,32 +438,25 @@
...
-                            </ContentControl>
+                            <Viewbox x:Name="PART_ActionIconPresenterHolder"
+                                     Grid.RowSpan="2"
+                                     Grid.Column="3"
+                                     Width="32"
+                                     Height="32"
+                                     MaxWidth="{ThemeResource SettingsCardActionIconMaxSize}"
+                                     MaxHeight="{ThemeResource SettingsCardActionIconMaxSize}"
+                                     Margin="{ThemeResource SettingsCardActionIconMargin}"
+                                     HorizontalAlignment="Center"
```
michael-hawker, line 449: Are any of these changes going to be a problem for folks updating? Would we see folks wanting to change the size here off 32x32? (smaller or bigger?)

This is the only part I'm trying to understand the nuances in of the PR in terms of ramifications of inverting the ViewBox/Content here.

Is this icon smaller than it used to be now? We don't have many icon examples in our sample app, so it's hard to tell.

image

The icon I think is just different on Win10 as well...

components/SettingsControls/src/SettingsCard/SettingsCard.Properties.cs
```diff
@@ -69,6 +68,15 @@ public partial class SettingsCard : ButtonBase
         typeof(SettingsCard),
         new PropertyMetadata(defaultValue: ContentAlignment.Right));
 
+    /// <summary>
+    /// The backing <see cref="DependencyProperty"/> for the <see cref="IsActionIconVisible"/> property.
+    /// </summary>
+    public static readonly DependencyProperty IsActionIconVisibleProperty = DependencyProperty.Register(
```
michael-hawker, line 74: Think name is fine, kind of maps to stuff like `NavigationView` has like [`IsPaneToggleButtonVisible`](https://learn.microsoft.com/en-us/windows/winui/api/microsoft.ui.xaml.controls.navigationview.ispanetogglebuttonvisible?view=winui-2.8).

components/SettingsControls/src/SettingsCard/SettingsCard.xaml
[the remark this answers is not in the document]
niels9001: image

Oh good catch! This seems to happen on the `SettingsExpander` only due to the internal `SettingsCard` having no right-padding and thus the focus visual gets clipped. Fixed now!

components/SettingsControls/src/SettingsCard/SettingsCard.xaml
niels9001: Oh yes, the `Height` and `Width` values are redundant here. Without those, it renders as expected on W11 (and changing the `SettingsCardActionIconMaxSize` still works to make icons smaller/bigger):
image

niels9001:
@michael-hawker Maybe only merge this in after #417 so that we have the version bump with this PR?

michael-hawker (approved):

michael-hawker:
@niels9001 updated now that 417 is in, so should be good to merge once CI passes. 🎉

PR #419, opened 2023-04-15
Fixes #388 SizerBase Orientation Property

### LINKED ISSUE #388: [SizerBase] Orientation and Cursor not set correctly
Parent Experiment: https://github.com/CommunityToolkit/Labs-Windows/issues/101

Figured out what was going on with the cursor not properly updating when the Orientation is Horizontal (like in the sample) - regression from current Toolkit version.

It's this code here:

https://github.com/CommunityToolkit/Labs-Windows/blob/9045236e77884e6c5c9fccb261e390001c93d4ec/labs/SizerBase/src/SizerBase.Properties.cs#L98-L110

Since we set the value of the dependency property, when the initial pass run on load sets it to `SizeWestEast`. Then the detection logic runs and `Orientation` is changed to `Horizontal`, but the set gets ignored the 2nd time as the prop has been set...

Think this was abstracted in the old code as the inner class held that info.

Here, I think we need to just have the private value tracking what the cursor should be and leave the Cursor property for overrides? Will think about it more and get back to it as a separate fix.

Effects UWP and WASM (basically WinUI 2), Windows App SDK uses a different setup, so I think is uneffected.

_Originally posted by @michael-hawker in https://github.com/CommunityToolkit/Labs-Windows/issues/101#issuecomment-1334651826_
            
(Will have to create a new issue template for experiment bugs...)

michael-hawker:
Fixes #388

- Updates to WASDK 1.3 (resolves quirky/jumpy behavior on WindowsAppSDK due to Manipulation event bug in platform - see https://github.com/microsoft/microsoft-ui-xaml/issues/6847)
- Removes Mouse Extension from XAML Template and does in code-behind as to remove the need to set the Cursor property directly via binding
  - By setting the Cursor DependencyProperty we were losing our detection mechanism, now if it's set it's an explicit override, otherwise we use our logic based on Orientation property
- Also, switches to new CommunityToolkit.*.Extensions Dependency vs. old copies of Tree helpers
- Fixes some doc typos
- Bumps version, tested locally on UWP, WASDK, and Uno.UI/WASM

Just need to see if anything breaks in the CI due to the update to the WASDK... Otherwise, should be good to go and will help with #418's next step. 🎉

Glad to finally get this stuff resolved, this has been in the back of my mind for a while now. The new SizerBase is now above-and-beyond what we currently have in WCT. Just needs some future modern magic to discuss with @niels9001 as the final frontier (as a separate PR in the future).


michael-hawker:
Looks like the extensions broke in the tests with the flip or something...

michael-hawker:
Hmm, WinUI 3 failed to build do to namespace conflict... Wonder if from global using I added in tooling? Or could be the other components references to the older package, trying to change that, though Shimmer uses the animation package which has an indirect dependency... so not sure if that's still going to be trouble. This is a bit complex here with the old vs. new incompatibility with the namespace change... hmm... Would make sense that the WinUI 2 build works here as it was a completely different root to the namespace anyway.

michael-hawker:
Think going to have to bring other helpers over, as there's the CompositionTargetHelper and WeakEventListener used in here too... hmmm

michael-hawker:
Thinking about it more, we probably want to wait in using the new unified packages until more of them are setup. Otherwise, they'll start forcing folks to also use them over the 7.1.x packages when trying to use Labs components, don't think we're quite ready for that...

Will merge Niels' new style PR in first after I test on Win 10 - https://github.com/michael-hawker/Labs-Windows/pull/2 - and then untease/revert back to the 7.x extension usage.

michael-hawker:
Now that https://github.com/CommunityToolkit/Windows/pull/26 is merged in the new repo, we can just close this, it has the same fix as here. Should probably remove SizerBase from this repo now.

PR #430, opened 2023-05-06
FIX ICONS

Jay-o-Way:
@niels9001 can we stop this game already?


niels9001:
> @niels9001 can we stop this game already?

Mistakes happen, no need for these comments in every topic.. Thanks for fixing it though!

niels9001 (approved):
Thanks! Will merge when CI checks pass!

Jay-o-Way:
Somebody in your position should not be making mistakes like this. I shouldn't be cleaning up after you.

michael-hawker:
@Jay-o-Way this isn't appropriate. We're all human, things happen. 

Appreciate you making a PR and correcting this issue here, but there's no reason to be rude.

PR #459, opened 2023-07-04
[Experiment] TitleBar

niels9001:
Addressing: https://github.com/CommunityToolkit/Labs-Windows/discussions/454

Titlebar


ghost1372:
Tnx @niels9001 

Some Issues:
1. TitleBar Buttons Should be updated based on windows theme in Active/DeActive State:



2. we can not use VerticalAlignment="Center" for AutoSuggestBox or other controls

image

ghost1372:
@niels9001 
For some of the problems I reported, I can create a PR. Is that possible?

niels9001:
> Tnx @niels9001 
> 
> Some Issues:
> 1. TitleBar Buttons Should be updated based on windows theme in Active/DeActive State: (i can fix this issue, can i make a pr for this?)
> 

Sure, go ahead - feel free to PR against this branch!



> 3. ContextMenu theme is wrong and always light (i can fix this issue, can i make a pr for this?)

I believe WASDK 1.4 should I address this issue.. let me sync with that team to see if that's correct and we don't need to do any registry stuff?

ghost1372:
i pushed a commit here
https://github.com/ghost1372/Labs-Windows/commits/UpdateCaptionButtons

I don't know if it is possible to merge another PR on this PR?
Anyway, I submited a PR here https://github.com/CommunityToolkit/Labs-Windows/pull/460

dongle-the-gadget:
> ContextMenu theme is wrong and always light (i can fix this issue, can i make a pr for this?)

IIRC it requires a private API to be used extremely early into the loading process.

Jay-o-Way:
> in TitleBar Height = Standard , we can not use VerticalAlignment=\"Center\" for AutoSuggestBox or other controls

That's the reason "tall" should be used as soon as there are interactive/input controls. Don't know if it's possible to force this setting, in the logic, based on content?

Jay-o-Way:
@ghost1372 you might want to re-count those pixels

ghost1372:
> you might want to re-count those pixels
I don't want the size of the buttons to be big, just like the store, I want it to be small, but the autosuggestbox stays in the center!
Screenshot 2023-07-11 185529

ghost1372:
@niels9001 
`AutoConfigureCustomTitleBar="False"
`
Screenshot 2023-07-11 185847

Jay-o-Way:
@ghost1372 oh sorry, now i see those buttons. @niels9001 those should stay untouched.

niels9001:
@Jay-o-Way Please leave any generic questions/comments in the discussion and not in this PR.

niels9001:
> 



> > > in TitleBar Height = Standard , we can not use VerticalAlignment="Center" for AutoSuggestBox or other controls
> > 
> > 
> > That's the reason "tall" should be used as soon as there are interactive/input controls. Don't know if it's possible to force this setting, in the logic, based on content?
> 
> Microsoft itself has used the standard mode, so we definitely do not need the tall mode to implement this mode, Store is a perfect example Screenshot 2023-07-11 183528
> 
> @niels9001 When will you merge this PR? It's been almost 2 weeks...

I have some other stuff that got prioritized. I think in terms of features, we are in a good spot with this PR - or do you see any must-have things that need to be fixed still? I think having some basic tests in place would be good to have - if that's something you want to help out with let me know!

ghost1372:
oh i got it, In general, everything is good, 
There are just a few minor issues:
1. `AutoConfigureCustomTitleBar="False"` It causes a null exception
2. `VerticalAlignment="Center"` for AutoSuggestBox or other controls
3. The title bar is very slow to apply and I can practically see the default title bar!
Animation

components/TitleBar/src/TitleBar.WASDK.cs
```diff
@@ -0,0 +1,208 @@
...
+            dragRectsList.Add(dragRectR);
+
+            Windows.Graphics.RectInt32[] dragRects = dragRectsList.ToArray();
+
+            Window.AppWindow.TitleBar.SetDragRectangles(dragRects);
+        }
+    }
+
+    [DllImport("Shcore.dll", SetLastError = true)]
+    internal static extern int GetDpiForMonitor(IntPtr hmonitor, Monitor_DPI_Type dpiType, out uint dpiX, out uint dpiY);
```
michael-hawker, line 180: Seems silly for a single method, but can use [CsWin32](https://github.com/microsoft/CsWin32) to generate these instead...

TyJOrtiz:
> @niels9001  oh i got it, In general, everything is good, 
> 
> There are just a few minor issues:
> 
> 1. `AutoConfigureCustomTitleBar="False"` It causes a null exception
> 
> 2. `VerticalAlignment="Center"` for AutoSuggestBox or other controls
> 
> 3. The title bar is very slow to apply and I can practically see the default title bar!
> 
> Animation
> 
> 
> 
> 
> 
> I have an app that has used a customized title bar (codes from MS documentation) and it is very fast and nothing is visible, but TitleBar Control is slow!
> 
> Animation2
> 
> 

I think that’s just a current limitation of WinUI 3. The Store App is UWP so it’s merely extending the view into the title bar instead of having to hide the title bar then load the control in WinUI 3.

ghost1372:
> > 
> > 
> 
> I think that’s just a current limitation of WinUI 3. The Store App is UWP so it’s merely extending the view into the title bar instead of having to hide the title bar then load the control in WinUI 3. 

See my second screenshot, it's a WinUI app, everything is normal!

niels9001:
> 
> > > 
> > > 
> > 
> > I think that’s just a current limitation of WinUI 3. The Store App is UWP so it’s merely extending the view into the title bar instead of having to hide the title bar then load the control in WinUI 3. 
> 
> See my second screenshot, it's a WinUI app, everything is normal!
> 
> @TyJOrtiz no its a winui 3 app

Ideally, the `ExtendContentIntoTitleBar` is set to true in `App.xaml.cs` - that's when you don't see it changing.

Doing that within the control does not seem to be esrly enough. Would it make sense to add this as guidance in the sample/doc? Or any other thoughts on how we could set this earlier?

ghost1372:
> Doing that within the control does not seem to be esrly enough. Would it make sense to add this as guidance in the sample/doc? Or any other thoughts on how we could set this earlier?

I do not have enough experience, I am currently investigating
This issue is mentioned in the Microsoft documentation


> Set ExtendsContentIntoTitleBar to true in the MainWindow constructor. If you set it in code that gets called later, the default system title bar might be shown first, and then hidden.

ghost1372:
@niels9001 I tested different ways and finally I came to a conclusion.
Unfortunately, I did not find a place to solve this issue in the title bar codes.
But we can solve this issue in the ` OnLaunched` method

The important thing is that we must write these codes after Window is **Created**, moving the code before or after `currentWindow.Activate();` has no effect

1. if we use `currentWindow.AppWindow.TitleBar.ExtendsContentIntoTitleBar` this is result: (Changing the background of the caption buttons is visible)

```
protected override void OnLaunched(Microsoft.UI.Xaml.LaunchActivatedEventArgs e)
        {
            currentWindow = new Window();
            currentWindow.AppWindow.TitleBar.ExtendsContentIntoTitleBar = true;
```



2. if we use `currentWindow.ExtendsContentIntoTitleBar` this is result: (everything is ok and normal)

```
protected override void OnLaunched(Microsoft.UI.Xaml.LaunchActivatedEventArgs e)
        {
            currentWindow = new Window();
            currentWindow.ExtendsContentIntoTitleBar = true;
```




It would be great if we could create the window in the `public App()` method, but unfortunately, the window is not created and always is null
```
public App()
{
    this.InitializeComponent();
    currentWindow = new Window(); // always is null
}
```

If this issue can be solved by the titlebar codes, it will be great, otherwise, this issue/solution should be mentioned in the documentation.

ghost1372:
@niels9001  I found an important bug!
If we do not use the **TitleBar.Content**, its location cannot be **dragged**!

ghost1372:
Founded Bugs:

1.If we do not use the **TitleBar.Content**, its location cannot be **dragged**!
Reason: `MinWidth="360"`     
```
<ContentPresenter x:Name="PART_ContentPresenter"
                                          Grid.Column="5"
                                          MinWidth="360"
```

image


2. PaneButton is not working some times:


3. in RTL layout Title, Icon and other buttons are spaced as far as the CaptionButtons (The first run everything is fine, when changing the size of the window, the controls are spaced out)

ghost1372:
Thank you @niels9001 
4 of the issues are fixed, but 2 of them remain

niels9001:
> Thank you @niels9001 4 of the issues are fixed, but 2 of them remain

Thanks for the great overview!!! Hopefully WASDK 1.4 resolves the RTL issue! Do you have any idea what is causing the menu button glitch? Almost feels like the dragregion is blocking it somehow?

ghost1372:
> > Thank you @niels9001 4 of the issues are fixed, but 2 of them remain
> 
> Thanks for the great overview!!! Hopefully WASDK 1.4 resolves the RTL issue! Do you have any idea what is causing the menu button glitch? Almost feels like the dragregion is blocking it somehow?

The issue occurs when the back button is shown or hidden, so when the back button changes state the drag region must be updated.

components/TitleBar/samples/SamplePages/ShellPage.xaml.cs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    private void NavView_Loaded(object sender, RoutedEventArgs e)
+    {
+        NavView.SelectedItem = NavView.MenuItems[0];
+    }
+
+    private void NavFrame_Navigated(object sender, NavigationEventArgs e)
+    {
+        appTitleBar.IsBackButtonVisible = NavFrame.CanGoBack;
```
Jay-o-Way, line 54: I think it would be good to call `SetDragRegionForCustomTitleBar()` here

components/TitleBar/samples/TitleBar.md
```diff
@@ -0,0 +1,28 @@
...
+---
+
+The `TitleBar` provides an easy way to display a modern titlebar experience. The control handles all the required APIs to extend content into the titlebar area and set custom drag regions. The control is set up in a way that it handles the correct design guidelines while being flexible in what content to show.
+
+> [!Sample TitleBarConfigSample]
+
+> [!Sample TitleBarFullSample]
+
+## Using TitleBar in WASDK apps
+If `AutoConfigureCustomTitleBar` is enabled, the `TitleBar` will make the required code-behind changes to make your content extend into the titlebar area and setting the correct caption background brushes. However, launching the app might briefly show the standard titlebar. To overcome this, make sure you set the following code in the `OnLaunched` method (in `App.xaml.cs`) or wherever you create your window:
```
Jay-o-Way, line 22:
```suggestion
If **AutoConfigureCustomTitleBar** is enabled, the **TitleBar** will make the required code-behind changes to make your content extend into the titlebar area and setting the correct caption background brushes. However, launching the app might briefly show the standard titlebar. To overcome this, set the following code in the **OnLaunched** method in _App.xaml.cs_ or wherever you create your window:
```

components/TitleBar/samples/TitleBar.md
```diff
@@ -0,0 +1,28 @@
...
+
+`
```
Jay-o-Way, line 24:
```suggestion
```CS
```

components/TitleBar/samples/TitleBar.md
```diff
@@ -0,0 +1,28 @@
...
+currentWindow.AppWindow.TitleBar.ExtendsContentIntoTitleBar = true;
+currentWindow.AppWindow.TitleBar.ButtonBackgroundColor = Microsoft.UI.Colors.Transparent;
+`
```
Jay-o-Way, line 27: \```

Jay-o-Way (commented):
Few tiny things. I suppose the (missing) newlines at end of files isn't such a big deal.

components/TitleBar/samples/SamplePages/ShellPage.xaml.cs
niels9001: The control is handling this already whenever this property is changed!

ghost1372:
@niels9001 
i found another bug!
we can not move window or exit full screen (drag & drop or double click on Titlebar) if we use Win + Right or Left


--

and this is MS Store which is working fine!
00000055555500000xxxx

ghost1372:
It seems that **SetDragRects** is marked as deprecated in the metadata and there is a new API in v1.4, 
`Microsoft.UI.Input.InputNonClientPointerSource.GetForWindowId(wndId).SetRegionRects(NonClientRegionKind, Rects)`

> InputNonClientPointerSource in the Microsoft.UI.Input namespace can be used for custom titlebar scenarios to define non-client area regions. Code can register for corresponding events like hover and click events on these regions.

Jay-o-Way:
https://github.com/microsoft/WindowsAppSDK/issues/3845
https://learn.microsoft.com/windows/windows-app-sdk/api/winrt/microsoft.ui.input.inputnonclientpointersource?view=windows-app-sdk-1.4

Starcloudsea:
> [microsoft/WindowsAppSDK#3845](https://github.com/microsoft/WindowsAppSDK/issues/3845) https://learn.microsoft.com/windows/windows-app-sdk/api/winrt/microsoft.ui.input.inputnonclientpointersource?view=windows-app-sdk-1.4

I see you quoted my question and I think you should take a look at this:
https://learn.microsoft.com/en-us/answers/questions/1362230/how-to-fix-appwindow-titlebar-setdragrectangles-no?comment=answer-1315251&page=1#comment-1370034
Now I'm confused, I don't know why others don't have problems, but I have problems... Can you help me?

Jay-o-Way:
I don't have any answer. Just thought I would share something related. All I know is that version 1.4 is very new, so I imagine many developers will struggle with the same thing. Eventually it's up to @niels9001, he has his connections. 👍🏻

Starcloudsea:
> I don't have any answer. Just thought I would share something related. All I know is that version 1.4 is very new, so I imagine many developers will struggle with the same thing. Eventually it's up to @niels9001, he has his connections. 👍🏻

Well, I'll look for the answer first.

ghost1372:
oh finally there is a sample code:
https://github.com/microsoft/WinUI-Gallery/pull/1360

niels9001:
> oh finally there is a sample code: [microsoft/WinUI-Gallery#1360](https://github.com/microsoft/WinUI-Gallery/pull/1360)

Oh, this is great! Should make things easier in for the WASDK code. WASDK 1.4 is required though, the Toolkit is still on 1.3.

pratikone:
> oh finally there is a sample code: [microsoft/WinUI-Gallery#1360](https://github.com/microsoft/WinUI-Gallery/pull/1360)

👋🏾 this PR is merged now and yes, it requires the use of new InputNonClient apis which are only available on winappsdk 1.4 onwards

AnalogFeelings:
It seems the title disappears when making the window narrow, even if there is no content in the middle?


https://github.com/CommunityToolkit/Labs-Windows/assets/51166756/7e710841-5bd3-447c-bd30-560d83f13db6

niels9001:
> @niels9001 there is a bug, when windows scale settings is 125%, and we use snap layout, drag regions is not working, i found out that we can use **Window.SizeChanged** and here we can call **SetDragRegionForCustomTitleBar();** and bug can be fixed. however we have a **Titlebar.SizeChanged** event but this can not fix this bug, so My question is, should the **Titlebar.SizeChanged** event be removed and we keep **Window.SizeChanged**? Or keep both?

What would you propose.. keeping both? Just want to make sure that these fixing are redundant when moving to WASDK 1.4? (I have no idea though, how much 1.4 changes things here!)

ghost1372:
> What would you propose.. keeping both? Just want to make sure that these fixing are redundant when moving to WASDK 1.4? (I have no idea though, how much 1.4 changes things here!)

I prefer to keep both. If you agree, I will send a pr.
I haven't used the new v1.4 API yet And I don't know about it.
But I compiled and tested the titlebar with v1.4.

Jay-o-Way:
Is this (getting) up to date with the latest `InputNonClientPointerSource` ?

ghost1372:
> Is this (getting) up to date with the latest `InputNonClientPointerSource` ?

Yes i added v.1.4 api and titlebar now using **InputNonClientPointerSource**

insomniachi:
What's blocking this?
I don't see a preview package yet.

ghost1372:
> What's blocking this?
> I don't see a preview package yet.

We should wait for #518

AnalogFeelings:
Are there plans for a TabView version of this?

ghost1372:
> Are there plans for a TabView version of this?

accourding to this https://github.com/CommunityToolkit/Labs-Windows/discussions/454
Not planned for version one.
Although, I don't have any image in mind of integrating the titlebar with the tabview.
Is there an example?

AnalogFeelings:
> > Are there plans for a TabView version of this?
> 
> accourding to this https://github.com/CommunityToolkit/Labs-Windows/discussions/454
> Not planned for version one.
> Although, I don't have any image in mind of integrating the titlebar with the tabview.
> Is there an example?

I was thinking of something like notepad

Jay-o-Way:
> Although, I don't have any image in mind of integrating the titlebar with the tabview.
> Is there an example?

Pretty much any tabbed window. Notepad, Explorer, browsers...

ghost1372:
Thank you @Jay-o-Way  @AnalogFeelings 
I would like to work on it😁

ghost1372:
Hi @pratikone 
Can you help us?
which one we should use?

 **Window.ExtendsContentIntoTitleBar + Window.SetTitleBar()** 
or
**Window.AppWindow.TitleBar.ExtendsContentIntoTitleBar**

insomniachi:
Is there anything i need to do while hiding the titlebar ? 
i referred this from msdocs to hide the title bar
```C#
public MainWindow()
{
    this.InitializeComponent();

    m_AppWindow = this.AppWindow;
    m_AppWindow.Changed += AppWindow_Changed;
}

private void AppWindow_Changed(AppWindow sender, AppWindowChangedEventArgs args)
{
    if (args.DidPresenterChange)
    {
        switch (sender.Presenter.Kind)
        {
            case AppWindowPresenterKind.CompactOverlay:
                // Compact overlay - hide custom title bar
                // and use the default system title bar instead.
                AppTitleBar.Visibility = Visibility.Collapsed;
                sender.TitleBar.ResetToDefault();
                break;

            case AppWindowPresenterKind.FullScreen:
                // Full screen - hide the custom title bar
                // and the default system title bar.
                AppTitleBar.Visibility = Visibility.Collapsed;
                sender.TitleBar.ExtendsContentIntoTitleBar = true;
                break;

            case AppWindowPresenterKind.Overlapped:
                // Normal - hide the system title bar
                // and use the custom title bar instead.
                AppTitleBar.Visibility = Visibility.Visible;
                sender.TitleBar.ExtendsContentIntoTitleBar = true;
                break;

            default:
                // Use the default system title bar.
                sender.TitleBar.ResetToDefault();
                break;
        }
    }
}
```
because my application is crashing while returning from fullscreen mode to Overlappedmode. (this is not always, but it randomly happens. i'm not able to identify why or when it happens.)
but i'm sure it's something wrong with the titlebar because when i don't use it, App never crashes.

insomniachi:
in my case, it'll never call reset to default because i'm only switching between Overlapped/Fullscreen, i'm that i'm only changing visibility

ghost1372:
> in my case, it'll never call reset to default because I'm only switching between Overlapped/Fullscreen, in that I'm only changing visibility

So you should set AutoConfigureCustomTitleBar = false when you are switching between Overlapped/Fullscreen

If you can please upload a sample so i can test it better.

insomniachi:
reproduceable sample
https://github.com/insomniachi/CTK_TitleBar_Crash
it will not crash when running from visual studio.
- publish unpackaged win-x64
- run exe
- there should be a media player element in the main page
- play video
- double click the media player element to go to Fullscreen.
- while playing double click again to come back to overlapped. it should crash, or repeat the Fullscreen and back steps a few times, it should eventually crash.

ghost1372:
Tnx @insomniachi 
i tested 100 times but i did not see any crash, mybe something related to MediaPlayerElement or WinUI Windowing iteself, so i suggest you report issue in Microsoft.UI.Xaml or WindowsAppSDK repo

insomniachi:
did you try from visual studio or published exe ?
it does not happen when running from visual studio
Recording 2024-01-10 094232

ghost1372:
> did you try from visual studio or published exe ?
> it does not happen when running from visual studio
> Recording 2024-01-10 094232
> 

I tried both visual studio and publish

ghost1372:
i found another bug
if we collapse NavigationView, then we choose an item, we can not select first item and act as a titlebar drag region

Screenshot 2024-03-15 153256

TyJOrtiz:
Is it possible to just adjust the pop up margin?

ghost1372:
> i found another bug if we collapse NavigationView, then we choose an item, we can not select first item and act as a titlebar drag region

This is an expected behavior!
https://github.com/microsoft/microsoft-ui-xaml/issues/9448#issuecomment-2010941425

Arlodotexe:
Looks like this is blocking the CI:
```
"D:\a\Labs-Windows\Labs-Windows\components\TitleBar\src\CommunityToolkit.WinUI.Controls.TitleBar.csproj" (default target) (21:67) ->
(CoreCompile target) -> 
  D:\a\Labs-Windows\Labs-Windows\components\TitleBar\src\TitleBar.Properties.cs(45,73): error CS1574: XML comment has cref attribute 'Display' that could not be resolved [D:\a\Labs-Windows\Labs-Windows\components\TitleBar\src\CommunityToolkit.WinUI.Controls.TitleBar.csproj::TargetFramework=net7.0-android33.0]
```

Arlodotexe (approved):
We'd like to close this PR so that iterations and fixes can be done separately, and so that people can start using the experiment from a package in the Labs feed instead of a PR feed.

Arlodotexe:
The current warning in the CI:
```
"D:\a\Labs-Windows\Labs-Windows\tooling\ProjectHeads\AllComponents\Tests.Uwp\CommunityToolkit.Tests.Uwp.csproj" (default target) (31:6) ->
(CoreCompile target) -> 
  D:\a\Labs-Windows\Labs-Windows\components\TitleBar\tests\ExampleTitleBarTestClass.cs(80,23): error CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread. [D:\a\Labs-Windows\Labs-Windows\tooling\ProjectHeads\AllComponents\Tests.Uwp\CommunityToolkit.Tests.Uwp.csproj]
  D:\a\Labs-Windows\Labs-Windows\components\TitleBar\tests\ExampleTitleBarTestClass.cs(88,23): error CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread. [D:\a\Labs-Windows\Labs-Windows\tooling\ProjectHeads\AllComponents\Tests.Uwp\CommunityToolkit.Tests.Uwp.csproj]
  D:\a\Labs-Windows\Labs-Windows\components\TitleBar\tests\ExampleTitleBarTestClass.cs(95,23): error CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread. [D:\a\Labs-Windows\Labs-Windows\tooling\ProjectHeads\AllComponents\Tests.Uwp\CommunityToolkit.Tests.Uwp.csproj]
```

PR #461, opened 2023-07-06
UpdateCaptionButtons

ghost1372:
Update Caption Buttons when theme is changing
https://github.com/CommunityToolkit/Labs-Windows/pull/459

Animation


ghost1372:
> Thanks!!
> 
> Will merge this in for now, but we might want to revise for WASDK1.4 as I believe this issue will be fixed there. We also need to consider using predefined brushes for `Foreground` colors to properly support HighContrast mode?

WASDK 1.4-preview 1 Recently released, I will try to test it
https://www.nuget.org/packages/Microsoft.WindowsAppSDK/1.4.230628000-preview1

PR #464, opened 2023-07-11
DataTable design tweaks

niels9001:
- Replacing the custom `HeaderedItemsControl` and `HeaderedTreeView` with the Toolkit equivalents.
- Tweaks to the default `DataColumn` style + samples these are more inline with design specifications.
- The `Background`, `BorderBrush`, `BorderThickness` and `CornerRadius` of the `DataColumn` can now be set for additional customization..

image

@michael-hawker the virtualization sample has some weird offsets.. not sure why though?


components/DataTable/src/Dependencies.props
```diff
@@ -17,12 +17,14 @@
...
         <PackageReference Include="CommunityToolkit.Uwp.Controls.Sizers" Version="8.0.0-beta.2"/>
+        <PackageReference Include="CommunityToolkit.Uwp.Controls.HeaderedControls" Version="8.0.0-beta.1"/>
     </ItemGroup>
 
     <!-- WinUI 3 / WinAppSdk / Uno -->
     <ItemGroup Condition="'$(IsWinAppSdk)' == 'true' OR ('$(IsUno)' == 'true' AND '$(WinUIMajorVersion)' == '3')">
         <PackageReference Include="CommunityToolkit.WinUI.Converters" Version="8.0.0-beta.1"/>
         <PackageReference Include="CommunityToolkit.WinUI.Extensions" Version="8.0.0-beta.1"/>
         <PackageReference Include="CommunityToolkit.WinUI.Controls.Sizers" Version="8.0.0-beta.2"/>
+        <PackageReference Include="CommunityToolkit.WinUI.Controls.HeaderedControls" Version="8.0.0-beta.1"/>
```
michael-hawker, line 28: The HeaderedControls dependency is only needed for the samples now, eh? So, it can go in the Samples Dependencies file instead of the source for the control. `DataTable` doesn't specifically reference the Toolkit components since they're all based on either `ItemsPresenter` or the call out to `TreeView` itself.

michael-hawker:
Ugh, I think the listview styling is different between Win 10/11? 😟

image

Virtualization is looking pretty weird on my box:

image

TreeTable is looking good.

components/DataTable/src/Dependencies.props
Arlodotexe: Seconded, if we only use HeaderedControls for the Sample, then we can move the PackageReference there.

michael-hawker: @niels9001 I made a commit for this.

niels9001:
Which sample is the first one? Will have a look at that.

Yeah the virtualized demo shows up like that for me too.. I don't know why that's happening though, did a quick check but couldn't find any (XAML) margins or paddings causing this?

michael-hawker:
On the base branch (UWP), DataTable Setup sample:

image

Virtualization:

image

(The extra spacing with the sizer is weird, but due to the ColumnSpacing... hmm) Maybe it's just compounded by the smaller sizer size in the new update?

components/DataTable/samples/DataTableVirtualizationSample.xaml
```diff
@@ -13,13 +13,12 @@
               VerticalAlignment="Top"
               ItemsSource="{x:Bind InventoryItems}">
         <ListView.Header>
-            <controls:DataTable Margin="12,0,0,0"
-                                ColumnSpacing="16">
+            <controls:DataTable ColumnSpacing="16">
```
michael-hawker, line 16: Maybe it's the missing margin? I lot of this is a bit fickle to line things up atm... 😟 Still not sure of a better approach between the different hosts yet. At least if we can have them setup in the samples to be decent it's a starting point.

Arlodotexe (approved):

Arlodotexe:
Couldn't find any obvious issues beyond those caught by @michael-hawker.

michael-hawker:
Kicked off a CI build to ensure I got the dependency/test stuff resolved (worked locally): https://github.com/CommunityToolkit/Labs-Windows/actions/runs/5579215230

michael-hawker:
Yeah, I think there's differences in the padding/margin between ListView on UWP/WASDK... so this may just not be perfect unless we find a style/resource based solution or something to have setups for each type of control per platform?

components/DataTable/samples/DataTableVirtualizationSample.xaml
michael-hawker: But then it's not the same for WASDK... 😭

michael-hawker (approved):
Think going to have to find an alternate solution for this somehow... Maybe we need some sort of 'provider' class which can provide defaults to the template based on the parent container control and the platform... Kind of like what NavigationView did or something... Still maybe need to figure out what those are, and not sure if the timing would be right to have it tied to the style (would be nice for overriding though)...

For now, we'll worry about it later. 🙂

PR #470, opened 2023-07-15
Add HasTitleBar, IsResizable, IsMaximizable, IsMinimizable and IsAlwaysOnTop

ghost1372:
Having some of these properties seems necessary.
we can reach the following results:

`IsMaximizable="False"`

`IsMinimizable="False"`

```
HasTitleBar="False"
IsResizable="False"
```

### DIFF
--- a/components/TitleBar/src/TitleBar.Properties.cs
+++ b/components/TitleBar/src/TitleBar.Properties.cs
@@ -62,6 +62,16 @@ public partial class TitleBar : Control
     /// The backing <see cref="DependencyProperty"/> for the <see cref="Window"/> property.
     /// </summary>
     public static readonly DependencyProperty WindowProperty = DependencyProperty.Register(nameof(Window), typeof(Window), typeof(TitleBar), new PropertyMetadata(null));
+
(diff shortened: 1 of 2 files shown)


niels9001:
@ghost1372 A few questions:

- Aren't these things more related to the window vs. the titlebar?
- Are these WASDK specific or are there UWP equivalents?

ghost1372:
1.Yes, these things are mostly related to Window. But due to the limitations of Window itself, we cannot use them in xaml and Because some of these properties affect the title bar, I think it is useful to have them.

2.I don't have experience with UWP, but since we are using **Window.AppWindow** and **(Microsoft.UI.Windowing.OverlappedPresenter)** for this, it is very likely that this cannot be done in UWP.It may be done through other APIs in UWP, I don't know!

itsWindows11:
I don't think disabling caption buttons in a UWP CoreWindow is possible without private/limited access APIs.

niels9001:
@ghost1372 As this PR introduces changes that are meant for the appwindow itself vs. the titlebar specifically I want to hold off on merging it: we want to make sure that most features of this control work on UWP and WASDK in the same way - and keep the control lightweight and simple.

Maybe we can integrate these changes in one of the titlebar samples so we can show devs how to do these kinds of things in WASDK?

ghost1372:
good idea tnx.

PR #480, opened 2023-07-23
[Experiment] Markdig based MarkdownTextBlock

nerocui:
This PR adds a new experiment: MarkdownTextBlock. It's a new take on the existing MarkdownTextBlock in the community toolkit. This implementation use the markdig library for parsing markdown syntax.

This initial implementation contains only early work. After this, the plan is to make it more generic, extensible and customizable.

screenshots
image
 
image


CosmicPredator (approved):
LGTM!!!

components/MarkdownTextBlock/src/TextElements/MyImage.cs
```diff
@@ -0,0 +1,158 @@
...
+                    byte[] data = await response.Content.ReadAsByteArrayAsync();
+                    // Create a BitmapImage for other supported formats
+                    BitmapImage bitmap = new BitmapImage();
+                    using (InMemoryRandomAccessStream stream = new InMemoryRandomAccessStream())
+                    {
+                        // Write the data to the stream
+                        await stream.WriteAsync(data.AsBuffer());
+                        stream.Seek(0);
+
+                        // Set the source of the BitmapImage
```
itsWindows11, line 133: I think you can directly read the response as a stream instead?

itsWindows11 (commented):
Also I guess might want to use ArrayPool and (ReadOnly)Span if possible here.

components/MarkdownTextBlock/src/Renderers/ObjectRenderers/HtmlBlockRenderer.cs
```diff
@@ -0,0 +1,30 @@
...
+            var lineText = line.Slice.ToString().Trim();
+            if (String.IsNullOrWhiteSpace(lineText))
+            {
+                continue;
+            }
+            stringBuilder.AppendLine(lineText);
+        }
+
+        var html = Regex.Replace(stringBuilder.ToString(), @"\t|\n|\r", "");
+        html = Regex.Replace(html, @"&nbsp;", " ");
```
itsWindows11, line 25: I guess you can precompile the regex? Can be done by using `RegexOptions.Compiled`.

components/MarkdownTextBlock/src/TextElements/MyImage.cs
CosmicPredator: > I think you can directly read the response as a stream instead then set it as an image source?

Some images are dynamic in nature. So I guess that won't work.

components/MarkdownTextBlock/samples/MarkdownTextBlockCustomSample.xaml.cs
```diff
@@ -0,0 +1,613 @@
...
+namespace MarkdownTextBlockExperiment.Samples;
+
+/// <summary>
+/// An example sample page of a custom control inheriting from Panel.
+/// </summary>
+[ToolkitSample(id: nameof(MarkdownTextBlockCustomSample), "Custom control", description: $"A sample for showing how to create and use a {nameof(CommunityToolkit.Labs.WinUI.MarkdownTextBlock)} custom control.")]
+public sealed partial class MarkdownTextBlockCustomSample : Page, INotifyPropertyChanged
+{
+    private MarkdownConfig _config;
+    private const string _markdown = @"
```
michael-hawker, line 16: You should be able to use the new raw string literal here so that you don't have to escape everything.

components/MarkdownTextBlock/src/TextElements/MyFlowDocument.cs
```diff
@@ -0,0 +1,67 @@
...
+    public void AddChild(IAddChild child)
+    {
+        TextElement element = child.TextElement;
+        if (element != null)
+        {
+#if !WINAPPSDK
+            if (element is Windows.UI.Xaml.Documents.Block block)
+#else
+            if (element is Microsoft.UI.Xaml.Documents.Block block)
+#endif
```
michael-hawker, line 51: You can define a using alias at the top for these types and use the base type here without the split. i.e. keep all the splits above like:

```cs
#if !WINAPPSDK
using Block = Windows.UI.Xaml.Documents.Block;
#else
using Block = Microsoft.UI.Xaml.Documents.Block;
#endif
```

Or even in the csproj like you did for the other global usings.

components/MarkdownTextBlock/src/Renderers/UWPRenderer.cs
```diff
@@ -0,0 +1,157 @@
...
+using CommunityToolkit.Labs.WinUI.MarkdownTextBlock.Renderers.ObjectRenderers.Inlines;
+using CommunityToolkit.Labs.WinUI.MarkdownTextBlock.Renderers.ObjectRenderers.Extensions;
+using CommunityToolkit.Labs.WinUI.MarkdownTextBlock.TextElements;
+using Markdig.Renderers;
+using Markdig.Syntax;
+using Markdig.Helpers;
+
+namespace CommunityToolkit.Labs.WinUI.MarkdownTextBlock.Renderers;
+
+public class UWPRenderer : RendererBase
```
michael-hawker, line 11: Wrapping head around all the connections still with this, but we probably at minimum would want to call this a 'WinUIRenderer' if not something else?

components/MarkdownTextBlock/src/MarkdownConfig.cs
```diff
@@ -0,0 +1,13 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace CommunityToolkit.Labs.WinUI.MarkdownTextBlock;
+
+public record MarkdownConfig
+{
+    public string? BaseUrl { get; set; }
+    public string? Markdown { get; set; }
```
michael-hawker, line 10: Would be good for the text to be its own `Text` dependency property on the control vs. backed into the config, as it may change more than the rest of these things which could be shared between multiple instances of the control.

components/MarkdownTextBlock/src/MarkdownTextBlock.xaml
```diff
@@ -0,0 +1,26 @@
...
+            <Setter.Value>
+                <ControlTemplate TargetType="labs:MarkdownTextBlock">
+                    <Grid Padding="{TemplateBinding Padding}"
+                          HorizontalAlignment="{TemplateBinding HorizontalAlignment}"
+                          VerticalAlignment="{TemplateBinding VerticalAlignment}"
+                          Background="{TemplateBinding Background}"
+                          BorderBrush="{TemplateBinding BorderBrush}"
+                          BorderThickness="{TemplateBinding BorderThickness}"
+                          CornerRadius="{TemplateBinding CornerRadius}">
+                        <Grid x:Name="MarkdownContainer" />
```
michael-hawker, line 20: Is there the need for the nested grids? Why can't the content go in the outer grid? Otherwise, could the inner one be a Border or ContentPresenter?

components/MarkdownTextBlock/src/MarkdownUIBuilder.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public static UIElement Build(MarkdownConfig config)
+    {
+        if (config == null) throw new ArgumentNullException(nameof(config));
+        var pipeline = new MarkdownPipelineBuilder()
+            .UseEmphasisExtras()
+            .UseAutoLinks()
+            .UseTaskLists()
+            .UsePipeTables()
+            .Build();
```
michael-hawker, line 21: This is MarkDig configuration stuff, eh? Later, we probably want to think about how we can both cache this (for reducing overhead of re-rendering) as well as exposing this for additional settings by a developer.

michael-hawker (commented):
Did a high-level pass, some initial comments. Some things we can do now, some things to think about for later.

michael-hawker:
Ah, also looks like XAML Styler needs to be run, @nerocui you can run the script in the root with `-Main` and it should clean things up.

components/MarkdownTextBlock/samples/MarkdownTextBlockCustomSample.xaml.cs
nerocui: The escapes are just for escaping markdown. So they are needed anyway.

components/MarkdownTextBlock/src/MarkdownUIBuilder.cs
nerocui: Yes agreed. There's a lot to think about in this area. I'll work on a dev design doc after this PR.

niels9001:
I see that all styling happens in code behind.. would it make sense to move these to separate styles defined in XAML and we load those, when needed, in code behind? Not sure if there would be any performance implications though?

That would allow a developer to easily override a style (like we do [here](https://github.com/CommunityToolkit/Tooling-Windows-Submodule/blob/13a022906610038c6448a57ae7b1a261318589ff/CommunityToolkit.App.Shared/Renderers/ToolkitDocumentationRenderer.xaml#L21) with the current `MarkdownTextBlock` for the Gallery itself).

components/MarkdownTextBlock/samples/MarkdownTextBlockCustomSample.xaml
```diff
@@ -15,6 +15,7 @@
         </Grid.ColumnDefinitions>
         <controls:MarkdownTextBlock
             Grid.Column="0"
+            Text="{x:Bind Text, Mode=OneWay}"
```
michael-hawker, line 18: Could just bind to `MarkdownTextBox.Text` directly? It's a bit weird for a UI class to implement `INotifyPropertyChanged`, and there's no real reason to do so here, so could just simplify the sample with a direct binding. The config isn't modified so it should be fine to have as the default, eh?

michael-hawker:
@nerocui I'll try and take this for a spin later this week. Any discussions you think we need to resolve before an initial merge?

When you have a chance, mind rebasing on the latest `main`?

You'll want to delete your `AdditionalAssemblyInfo.cs` file and remove the `<Version>` tag in your src csproj file.

nerocui:
@michael-hawker I'll address the style in XAML issue later this week, but other than that I don't see anything else for the initial one.

components/MarkdownTextBlock/samples/MarkdownTextBlockCustomSample.xaml
nerocui: Addressed in latet version

components/MarkdownTextBlock/src/MarkdownTextBlock.xaml
nerocui: Removed in the latest version

components/MarkdownTextBlock/src/MarkdownConfig.cs
nerocui: Addressed in the latest version

components/MarkdownTextBlock/src/Renderers/UWPRenderer.cs
nerocui: Addressed in latest version

components/MarkdownTextBlock/src/TextElements/MyFlowDocument.cs
nerocui: Addressed

components/MarkdownTextBlock/src/Renderers/ObjectRenderers/HtmlBlockRenderer.cs
nerocui: addressed

components/MarkdownTextBlock/src/TextElements/MyImage.cs
nerocui: no going to change this one for now.

nerocui:
@michael-hawker the latest commit addressed all the comments so far. Please feel free to take another look.

michael-hawker:
CI Error:

```
D:\a\Labs-Windows\Labs-Windows\components\MarkdownTextBlock\src\TextElements\MyList.cs(60,34): 
error CS8524: The switch expression does not handle some values of its input type (it is not exhaustive) involving an unnamed enum value. For example, the pattern '(CommunityToolkit.Labs.WinUI.MarkdownTextBlock.TextElements.BulletType)6' is not covered. [D:\a\Labs-Windows\Labs-Windows\components\MarkdownTextBlock\src\CommunityToolkit.WinUI.Controls.MarkdownTextBlock.csproj]
```

nerocui:
@michael-hawker I think that works. I haven't had time to work on this, but I can spend some time to refresh the branch and get to a clean build.

nerocui:
Refreshed and it works locally. It's getting conflict in the submodule. I'm not sure how to resolve that.

niels9001:
> Refreshed and it works locally. It's getting conflict in the submodule. I'm not sure how to resolve that.

Could you try to update from main? Hopefully that would resolve the conflict

nerocui:
@niels9001 I rebased the whole PR from main. Will try again.

niels9001:
@Arlodotexe can you help out with the conflict?

@nerocui I think the tooling update isn't required for this PR to work, right?

nerocui:
@niels9001 That's right. Tooling is not part of this change.

nerocui:
@michael-hawker @niels9001 Rebased again with the conflicts in tools resolved

nerocui:
Could we get this reviewed and merged please?

nerocui:
@michael-hawker @niels9001 Gentle ping

michael-hawker:
Thanks @nerocui - kicked off the build, we'll see if your latest changes gets us a clean one. Think that's the main thing we've been looking for. This and the ColorPicker/TabbedCommandBar from the main repo are our main PR priorities at the moment.

michael-hawker:
@nerocui looks like there are still some build errors in the CI. One seems related to the tests, probably just needs a tweak to a name or something to just do a basic load of the component as a litmus test. The other seems maybe to have to do with ColorCode as a dependency?

```
 D:\a\Labs-Windows\Labs-Windows\tooling\ProjectHeads\AllComponents\Tests.Uwp\obj\x64\Release\ExampleMarkdownTextBlockTestPage.g.i.cs(18,57): error CS0234: The type or namespace name 'MarkdownTextBlock_ClassicBinding' does not exist in the namespace 'CommunityToolkit.WinUI.Controls' (are you missing an assembly reference?) [D:\a\Labs-Windows\Labs-Windows\tooling\ProjectHeads\AllComponents\Tests.Uwp\CommunityToolkit.Tests.Uwp.csproj]


"D:\a\Labs-Windows\Labs-Windows\tooling\ProjectHeads\AllComponents\Tests.Uwp\CommunityToolkit.Tests.Uwp.csproj" (default target) (29:6) ->
(_OnXamlPreCompileError target) -> 
  C:\Program Files (x86)\Windows Kits\10\bin\10.0.19041.0\XamlCompiler\Microsoft.Windows.UI.Xaml.Common.targets(486,5): error MSB4181: The "CompileXaml" task returned false but did not log an error. [D:\a\Labs-Windows\Labs-Windows\tooling\ProjectHeads\AllComponents\Tests.Uwp\CommunityToolkit.Tests.Uwp.csproj]


"D:\a\Labs-Windows\Labs-Windows\tooling\ProjectHeads\AllComponents\Uwp\CommunityToolkit.App.Uwp.csproj" (default target) (30:6) ->
(CoreCompile target) -> 
  D:\a\Labs-Windows\Labs-Windows\tooling\ProjectHeads\AllComponents\Uwp\obj\x64\Release\XamlTypeInfo.g.cs(3372,25): error CS0012: The type 'StyleDictionary' is defined in an assembly that is not referenced. You must add a reference to assembly 'ColorCode.Core, Version=2.0.0.0, Culture=neutral, PublicKeyToken=null'. [D:\a\Labs-Windows\Labs-Windows\tooling\ProjectHeads\AllComponents\Uwp\CommunityToolkit.App.Uwp.csproj]
  D:\a\Labs-Windows\Labs-Windows\tooling\ProjectHeads\AllComponents\Uwp\obj\x64\Release\XamlTypeInfo.g.cs(3377,18): error CS0012: The type 'StyleDictionary' is defined in an assembly that is not referenced. You must add a reference to assembly 'ColorCode.Core, Version=2.0.0.0, Culture=neutral, PublicKeyToken=null'. [D:\a\Labs-Windows\Labs-Windows\tooling\ProjectHeads\AllComponents\Uwp\CommunityToolkit.App.Uwp.csproj]
```

components/MarkdownTextBlock/src/TextElements/MyCodeBlock.cs
```diff
@@ -0,0 +1,88 @@
...
+
+    public MyCodeBlock(CodeBlock codeBlock, MarkdownConfig config)
+    {
+        _codeBlock = codeBlock;
+        _config = config;
+        _paragraph = new Paragraph();
+        var container = new InlineUIContainer();
+        var border = new Border();
+        border.Background = (Brush)Application.Current.Resources["ExpanderHeaderBackground"];
+        border.Padding = _config.Themes.Padding;
```
Arlodotexe, line 28: We should move assignment of these properties to the template binding.

nerocui:
@michael-hawker Fixed the test failure and updated the md file to include new icon and "Labs" label. For the reference error, I wasn't sure how to fix it. Do I need to start a separate PR in the tooling repo to add colorcode?

michael-hawker:
@nerocui that's weird, you define the dependency in the dependencies.props so that's all that should be needed... Do you know if you reference the `StyleDictionary` anywhere?

@Arlodotexe any thoughts?

nerocui:
@michael-hawker I added ColorCode.Core to dependencies.props, so hope that fixes things.

michael-hawker:
@nerocui you already have that defined in your csproj as it's a net standard package, so it shouldn't have to be in the dependencies.props, that file is only for UWP/WinAppSDK/Uno differentiation. So we can put that back. There's something else going on, but not too sure what's going on. @Arlodotexe is going to take a look.

JoeTomkinson:
> @michael-hawker I added ColorCode.Core to dependencies.props, so hope that fixes things.

No luck, looks as though it's still failing on depenency-related errors to ColorCode.Core

Arlodotexe:
I believe I've found the source of the issue. The reference to Microsoft.Toolkit.Uwp.UI.Controls.Markdown 7.1.2 on the UWP head contains a transitive reference to ColorCode.Core that's conflicting with the version being referenced in this component. 

Removing the reference to the 7.x MarkdownTextBlock control and cleaning all references to it in the head, the build error is resolved. Only reproduces in Release mode, not Debug.

@michael-hawker We'll have to upgrade the heads in tooling to use the Labs version of `MarkdownTextBlock` over the 7.x version to resolve the build error.

nerocui:
> @michael-hawker We'll have to upgrade the heads in tooling to use the Labs version of `MarkdownTextBlock` over the 7.x version to resolve the build error.

Isn't that kind of chicken and egg problem? We'd need to merge this PR to have tooling to reference to.

michael-hawker:
Ah, thanks for the insight @Arlodotexe. Another case of Toolkit building toolkit strikes again... any suggestions for a path for us to untangle that? Do we have to do an intermittent PR to the tooling submodule to disable the Markdown stuff or should we create a branch for the tooling here to get a clean CI pass all together and then just snap them together?

nerocui:
What if in this PR I don't include ColorCode.Core? I'll just add it back in once this one merges and tooling is able to reference the new control.

Arlodotexe:
@michael-hawker Considering that updates to the tooling will affect both the [main repo](https://github.com/CommunityToolkit/Windows) and Labs, we should probably create a new branch in Tooling for this PR. We can set a `ToolkitMarkdownTextBlockSourceProject` in Build.Directory.props like we did [here](https://github.com/CommunityToolkit/Windows/blob/9f0f7247443bb598996166f4389fe99c7569098b/Directory.Build.props#L10-L20), and then set it up [here](https://github.com/CommunityToolkit/Tooling-Windows-Submodule/blob/03d01befe8baf75fb5684ce639acf7ad5584f9d5/ProjectHeads/App.Head.props#L58-L69) so it can be consumed by the app head.

That will fix our CI issue here, but we won't be able to merge the tooling PR (or this one) into `main` until we have a NuGet package on the PR feed ready to be used in the CommunityToolkit/Windows repo for the other code path. When we've got that sorted and can close this PR, we'll update tooling once more to use the package from the public NuGet feed instead of the PR feed.

michael-hawker:
> What if in this PR I don't include ColorCode.Core? I'll just add it back in once this one merges and tooling is able to reference the new control.

@nerocui that's not a bad idea. Would it be hard to isolate? @Arlodotexe sound good?

nerocui:
> > What if in this PR I don't include ColorCode.Core? I'll just add it back in once this one merges and tooling is able to reference the new control.
> 
> @nerocui that's not a bad idea. Would it be hard to isolate? @Arlodotexe sound good?

I'll take a look tonight, but the usage of ColorCode is very limited. So it should be easy.

nerocui:
Updated with colorcode removed. The code block won't render without it. I don't think we need to work around it for now since we are adding it right back after this right?

niels9001:
Woohoo! Passed CI

components/MarkdownTextBlock/src/CommunityToolkit.WinUI.Controls.MarkdownTextBlock.csproj
```diff
@@ -0,0 +1,43 @@
...
+    <ToolkitComponentName>MarkdownTextBlock</ToolkitComponentName>
+    <Description>This package contains MarkdownTextBlock.</Description>
+
+    <!-- Rns suffix is required for namespaces shared across projects. See https://github.com/CommunityToolkit/Labs-Windows/issues/152 -->
+    <RootNamespace>CommunityToolkit.WinUI.Controls.MarkdownTextBlockRns</RootNamespace>
+    <ImplicitUsings>enable</ImplicitUsings>
+  </PropertyGroup>
+
+  <!-- Sets this up as a toolkit component's source project -->
+  <Import Project="$(ToolingDirectory)\ToolkitComponent.SourceProject.props" />
```
Arlodotexe, line 12:
```suggestion
  <Import Project="$(ToolingDirectory)\ToolkitComponent.SourceProject.props" />

   <PropertyGroup>
    <PackageId>$(PackageIdPrefix).$(PackageIdVariant).Controls.$(ToolkitComponentName)</PackageId>
  </PropertyGroup>
```

Arlodotexe (approved):

nerocui:
@michael-hawker shall we merge?

Balkoth:
Will this be brought forward to the next release?

michael-hawker:
@Balkoth right now this is just a labs component, it'll remain here until it gathers enough support/polish to be considered ready for release.

We should have a tracking issue to outline what phases/status this component is in, but it looks like we forgot to get that added before we approved and merged this. @Arlodotexe I think there may be a discussion, we should ensure we have both created or labeled properly and linked to in the markdown tags for the component. Though we should also evaluate how to best use the new tasklist/sub-issue stuff GitHub has now too to track work/issues needed for components in those parent issues.

Balkoth:
This was already a component of the previous toolkit and not bringing it forward is an issue for me. Thats why i asked.

michael-hawker:
@Balkoth this is the intended replacement. We've just been pretty focused on supporting Native AOT at the moment for our upcoming release. If you encounter issues trying to upgrade to the Labs Markdown component or a scenario not yet supported by this new one, please let us know. Feel free to file issues, suggest solutions, and open a PR. Contributions are welcome. Thanks!

PR #481, opened 2023-07-25
Updated tooling to latest main

Arlodotexe:
In a dedicated PR to catch potential build errors.


Arlodotexe:
Will need this patch as well - https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/113

Arlodotexe:
Investigation into CI errors is continuing here - https://github.com/CommunityToolkit/Windows/pull/157

michael-hawker:
@Arlodotexe updated this PR/branch based on https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/116 and made the other required changes.

We'll see if this works now, big test... 🤞

This does mean with the automated versioning each PR will update all the packages, will need to look at the https://github.com/marketplace/actions/changed-files tool to look at only pushing packages with changes.

michael-hawker (approved):
Alright, hopefully we're all in-sync now.

Appears the root of the issue was the change made to the FrameworkElment AncestorType method signature based on a proposed bug fix. As it should only attach to `FrameworkElement` and not `DependencyProperty`.

However, this caused a signature mismatch and nuget resolution was pulling the prior packages still until we've re-aligned here. Hopefully, we won't hit this in the future again... but at least have some better ideas about what happened and how it can manifest, though it wasn't the easiest to catch at first, the output from the CI when the test app crashes is fairly non-existent. I had a lot of issues getting logging to work, and upgrading to the latest test code is blocked on WASDK, see https://github.com/microsoft/testfx/issues/1726. So future improvements are needed later.

PR #491, opened 2023-08-22
[Experiment] OpacityMaskView

h82258652:
Addressing: #490 
image


niels9001:
@h82258652 Please run `ApplyXamlStyling.ps1` in the root folder to apply Xaml Styler!

components/OpacityMaskView/src/OpacityMaskView.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        _maskBrush = _compositor.CreateMaskBrush();
+        _maskBrush.Source = GetVisualBrush(contentPresenter);
+        _mask = GetVisualBrush(maskRectangle);
+        _maskBrush.Mask = OpacityMask is null ? null : _mask;
+
+        SpriteVisual redirectVisual = _compositor.CreateSpriteVisual();
+        redirectVisual.RelativeSizeAdjustment = Vector2.One;
+        redirectVisual.Brush = _maskBrush;
+        ElementCompositionPreview.SetElementChildVisual(rootGrid, redirectVisual);
```
michael-hawker, line 74: Would it be possible to apply any of this like we do for the `AttachedCardShadow`/`AttachedShadow` or `UIElementExtensions.VisualFactory` components in the Toolkit, so it's just an extension that can be applied to any element directly vs. a ContentControl?

h82258652: @michael-hawker I don't think we can do this. Sergio created one like `UIElementExtensions.OpacityMask` in store, but the position of the `redirectVisual` is not aligned with the source control at some times. When this happens, we use a `Border` to wrap the content to fix. I think the one in the toolkit should be more generic.

michael-hawker: Sounds similar to what we have to do for the `AttachedDropShadow` at times. I think having specific requirements can be fine, it's just extensions are a lot easier to add into a design than having to add another layer of depth to the Visual Tree to all cases and setup the layout. It's probably a case of a templated control vs. a raw-element with the CardShadow too I'd imagine. What are your thoughts @Sergio0694?

components/OpacityMaskView/samples/Assets/icon.png
michael-hawker, line 1: @niels9001 / @Arlodotexe just need latest icon here.

h82258652: @michael-hawker updated.

components/OpacityMaskView/samples/OpacityMaskViewSample.xaml
```diff
@@ -0,0 +1,45 @@
...
+                    </Rectangle.Fill>
+                </Rectangle>
+            </controls:OpacityMaskView.OpacityMask>
+            <Button Content="Hello Windows Community Toolkit" />
+        </controls:OpacityMaskView>
+        <controls:OpacityMaskView>
+            <controls:OpacityMaskView.OpacityMask>
+                <Rectangle>
+                    <Rectangle.Fill>
+                        <LinearGradientBrush StartPoint="0,0" EndPoint="1,1">
```
niels9001, line 29: @h82258652 Would it make sense to add a `RadialGradientBrush` example here as well?

h82258652: Added.
image

components/OpacityMaskView/samples/OpacityMaskView.md
```diff
@@ -0,0 +1,18 @@
...
+keywords: OpacityMaskView, Control, Layout
+dev_langs:
+  - csharp
+category: Controls
+subcategory: Layout
+discussion-id: 490
+issue-id: 0
+icon: assets/icon.png
+---
+# OpacityMaskView
```
niels9001, line 14:
```suggestion
```

niels9001 (approved):
Tested this in the AI Dev Gallery, works great!

Arlodotexe (approved):

PR #496, opened 2023-09-02
Unblock DataTable on Uno

Martin1994:
This PR unblocks using DataTable on Uno with a Release build.

The top issue for now is the lack of ListView header in Uno: https://github.com/unoplatform/uno/issues/1979, so this patch introduced a new container without use of a header to unblock its usage on Uno.

This approach is kinda hacky though, so I am looking for discussion about the long term direction assuming ListView's header won't be implemented in Uno soon.

Tested with WinUI 3 and Uno on Skia (WPF and GTK).

data-table


Martin1994:
@dotnet-policy-service agree

michael-hawker:
Thanks @Martin1994. But yeah, I was trying to explicitly not have an extra layer/type here and just use the natural container found in these scenarios. I believe the Uno folks are working on the issue though with ItemsPresenter? Maybe @MartinZikmund can give an update?

Martin1994:
>  I believe the Uno folks are working on the issue though with ItemsPresenter?

That's great!

If header won't be a blocker, shall we take a look at the change at column resize handler?

```
// old
CurrentWidth = new(this.ActualWidth);
// new
CurrentWidth = new(this.Width);
```

When a manipulation event is sent on Uno, I don't see `ActualWidth` being changed so as a result the column will be frozen at its initial dimension.

MartinZikmund:
@michael-hawker @Martin1994 On our side @ramezgerges is now working on adding Header and Footer support here https://github.com/unoplatform/uno/pull/13507

michael-hawker:
Looks like the PR was merged and the issue closed: https://github.com/unoplatform/uno/issues/10745

Once @Arlodotexe gets Labs up-to-date with our latest changes from the Windows/tooling repos, then we can try the latest Uno version and see how things work with DataTable

Arlodotexe:
@MartinZikmund You mentioned that this should be fixed under the latest Uno package? If I understand, that won't require the changes in this PR if our package is updated, right?

MartinZikmund:
@Arlodotexe yes, changes to the PR should not be needed

Arlodotexe:
Great, thanks @MartinZikmund! Did some testing, looks like DataTable still has a few issues to resolve. I've moved tracking [here](https://github.com/CommunityToolkit/Labs-Windows/issues/559).

PR #485, opened 2023-08-02
Upgrade to .NET 7, enable WASM optimizations

Arlodotexe:
Companion PR to https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/117, which should be closed before merging this.

As noted:
> Any codebase consuming the tooling repo will need to be adjusted to account for the new nullable annotations and analyzers added between netstandard2.0 and net7.0.

This PR pulls in the .NET 7 upgrade and fixes the new warnings and errors that appeared in our codebase as a result.

PR #503, opened 2023-10-26
Fixed missing nuget source config

Arlodotexe:
Follow-up hotfix for an issue introduced in https://github.com/CommunityToolkit/Labs-Windows/pull/485

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -152,17 +152,12 @@ jobs:
         run: msbuild.exe CommunityToolkit.AllComponents.sln /restore /nowarn:MSB4011 -p:Configuration=Release 
 
       # Build All Packages
-      - name: pack experiments
+      - name: Pack experiments
         working-directory: ./tooling/Scripts/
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -156,6 +156,11 @@ jobs:
         working-directory: ./tooling/Scripts/
         run: ./PackEachExperiment.ps1 -date ${{ env.VERSION_DATE }}${{ env.VERSION_PROPERTY != '' && format(' -postfix {0}', env.VERSION_PROPERTY) || '' }}
         
+      # Push Packages to our DevOps Artifacts Feed
+      - name: Add source
+        if: ${{ github.ref == 'refs/heads/main' }}
+        run: dotnet nuget update source LabsFeed --username dummy --password ${{ secrets.DEVOPS_PACKAGE_PUSH_TOKEN }}
+
```
michael-hawker, line 163: I would suggest adding this step after the next one line 172 (as displayed here), so it's the 'Add source' and then 'Push packages' steps that are related right next to each other?

michael-hawker: We could also just collapse these into a single step anyway, as there'll always be run together, that may be the easiest, just use the same format as the PR version.

michael-hawker (approved):

PR #508, opened 2023-11-06
Use WASDK v1.4 new Custom TitleBar API, Simplify Calculating Rects and Fix TitleBar Drag Region for Scale 125%

ghost1372:
## Fix drag window
This PR Fixed title bar drag issue for scale 125%.

## Move To v1.4 API
This PR uses the new Custom TitleBar API (in wasdk v1.4). It also simplifies the logic for calculating the rectangles for the title bar

## Add Support for Fluent ContextMenu (MenuFlyout)
if user did not define TitleBar.ContextFlyout titlebar will be use default system context menu and if user define a menuflyout in TitleBar.ContextFlyout, default system context menu will be gone and menu flyout will be shown in titlebar.

Screenshot 2023-11-13 231517


ghost1372:
@niels9001 Can you check this pr? i tested on my laptop and everythings is working great.

ghost1372:
@niels9001 
which one we should use?

`Window.AppWindow.TitleBar.RightInset / scale
`
Screenshot 2023-11-11 211954


without scale:

Screenshot 2023-11-11 212151

ghost1372:
@niels9001 it seems that we should use Window.Extend + Window.SetTitleBar() instead of using Window.AppWindow.TitleBar.Extend...
I searched a bit and everyone seems to say that the first one should be used.
what is your opinion? Which one should we use?

ghost1372:
i found a way to add context menu for titlebar 😁
i will push a commit when i done

ghost1372:
@niels9001  can you answer my questions? tnx

ghost1372:
Thank you @niels9001 

please merge it🙏

PR #514, opened 2023-11-30
[DataTable] Update to release dependencies and fix Padding value for Sizer in DataColumn

michael-hawker:
Simple update to fix issue with DataColumn sizer display (and update to the latest dependencies)

Before, Sizer was invisible, could tell it was there from hover effect though:
image

After, Sizer appears again:
image

Tested on UWP and WASDK

Forgot that I don't have to update a Version field anymore, woot! 🎉 (though guess it means it respits out all the packages still...)

Note requires update to sub-module as we weren't on latest package versions of shipped 8.0. Think there was an issue/discussion mentioning this somewhere as well? FYI @Arlodotexe

### DIFF
--- a/components/CanvasView/src/Dependencies.props
+++ b/components/CanvasView/src/Dependencies.props
@@ -11,21 +11,21 @@
 <Project>
   <!-- WinUI 2 / UWP -->
   <ItemGroup Condition="'$(IsUwp)' == 'true'">
-    <PackageReference Include="CommunityToolkit.Uwp.Helpers" Version="8.0.230801-preview"/>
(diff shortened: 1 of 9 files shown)


niels9001 (approved):

michael-hawker:
Thanks @niels9001, @Arlodotexe mind looking at the dependent https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/159 and then I can update the pointer here after to main.

Arlodotexe (approved):
Looks good, just need to update with the latest from tooling

michael-hawker:
Alright, updated to the latest tooling main now that that's been merged.

PR #516, opened 2023-12-15
Ci-run/release-day

FinesseAC:

PR #517, opened 2023-12-15
Update launch.json

FinesseAC:


Arlodotexe:
Same as #516, we plan on phasing out usage of this extension. See https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/71.

PR #522, opened 2024-01-26
Update AppServices Microsoft.Windows.SDK.Contracts package to 10.0.22621.2428

michael-hawker:
@dpaulino was seeing an issue trying to use the AppServices library from a UWP app:

```
2>C:\Program Files (x86)\Windows Kits\10\bin\10.0.22621.0\XamlCompiler\Microsoft.Windows.UI.Xaml.Common.targets(410,5): XamlCompiler error WMC0901: A type with a full name of 'Windows.UI.Xaml.DependencyObject' already exists in an included reference.
```

I'm not too sure what happened, but seemed familiar to issues I had faced updating to the latest WASDK when the contracts and stuff weren't also updated alongside. Wonder if some of the updates to our build/tooling have caused a requirement shift here?

Anyway, I tried forcing an update to the project for the `Microsoft.Windows.SDK.Contracts` package to the latest stable `10.0.22621.2428` and that seemed to resolve the build issue.

Figure updating that in the package here should hopefully resolve the issue without having to declare it in the app?

@dpaulino once this builds, you should be able to try the PR feed one and see if it resolves the issue directly.


dpaulino:
@Sergio0694 you mentioned that you found the root cause of the issue. What other changes does this pr need to fix it?

Sergio0694:
Yeah I don't think this PR is not going to fix the issue. The NuGet packaging got messed up:

image

You can see we're completely missing the UWP TFM. For comparison, here's what's in 0.0.3 we're using in the Store:

image

Crucially, the UWP TFM doesn't have the package reference to the contracts at all (as it already has WinRT references).

michael-hawker:
@Arlodotexe what happened here, why is the UWP target missing?

dpaulino:
What needs to be done for this to move forward?

michael-hawker:
> What needs to be done for this to move forward?

@dpaulino we've been hung up migrating to .NET 8, that looks like it may be resolved now, and we're still waiting on our other blocking issue for ColorPicker, so Arlo should be able to take a look at this soon and what might have changed here.

Arlodotexe:
@Sergio0694 I did a local build and pack of AppServices, it doesn't look like any TFMs are missing at least. 

image

I'll run a test to make sure it works as expected in a blank UWP app.

Sergio0694:
Yeah that looks good! Does this mean there's something wrong just with the CI build? 🤔

Arlodotexe:
> Yeah that looks good! Does this mean there's something wrong just with the CI build? 🤔

Here is yesterday's prerelease build pushed to the Labs feed. 

image

Yes, it looks like this might be an issue in the CI. I'll poke around and see what I can find. 

@Sergio0694 Since this takes a bit of setup in a blank app, do you have a template on hand that you use for testing? Maybe the `HelloAppServices` project from the [announcement post](https://devblogs.microsoft.com/ifdef-windows/microsoft-store-open-source-series-appservices-library/)?

Arlodotexe:
Looks like this is fixed in the most recent most available in the Labs feed:

image

PR #525, opened 2024-02-12
[Shimmer] Invert dark theme colors

citelao:
Currently, the shimmers are difficult to see in dark mode. This change makes them easier to see.

I believe this is because the colors got transposed: for *light mode*, the **more translucent** color is used for 2 & 3, whereas currently for light mode, the **less translucent** color is used for 2 & 3. My _hunch_ is that this intended to account for the inversion in color b/w dark & light mode, but that's already accounted for because the colors themselves are inverted b/w those modes.

In other words, both the colors _and_ the translucencies were transposed, when only one should have been.

Before:

DarkShimmerBefore

After:

DarkShimmerAfter

## What changed?

* Swapped the translucencies in the dark mode colors.

## How tested?

Tested using our C++ port of this code.

### DIFF
--- a/components/Shimmer/src/Shimmer/Shimmer.cs
+++ b/components/Shimmer/src/Shimmer/Shimmer.cs
@@ -159,10 +159,10 @@ public partial class Shimmer : Control
         {
             case ElementTheme.Default:
             case ElementTheme.Dark:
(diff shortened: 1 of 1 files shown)


niels9001:
@citelao Thanks for the PR, added the before vs after images

niels9001 (approved):
LGTM!

citelao:
Thanks @niels9001! The screenshots really help sell it :)

PR #528, opened 2024-02-28
Ensure VERSION_PROPERTY is in .nupkg name when provided

Arlodotexe:
This should help us avoid issues like https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/171 in the future

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -158,6 +158,10 @@ jobs:
         working-directory: ./tooling/Scripts/
         run: ./PackEachExperiment.ps1 -date ${{ env.VERSION_DATE }}${{ env.VERSION_PROPERTY != '' && format(' -postfix {0}', env.VERSION_PROPERTY) || '' }}
 
+      - name: Validate package names
+        if: ${{ env.VERSION_PROPERTY != '' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -158,6 +158,10 @@ jobs:
         working-directory: ./tooling/Scripts/
         run: ./PackEachExperiment.ps1 -date ${{ env.VERSION_DATE }}${{ env.VERSION_PROPERTY != '' && format(' -postfix {0}', env.VERSION_PROPERTY) || '' }}
 
+      - name: Validate package names
+        if: ${{ env.VERSION_PROPERTY != '' }}
```
michael-hawker, line 162: Should we always run this? Or throw earlier above if this env isn't set? As we should always expect it to be set, eh?

Arlodotexe: This check is isn't to make sure `VERSION_PROPERTY` doesn't go missing it's to make sure any nuget packages produced have the version property if it's already set. The `VERSION_PROPERTY` is always set in labs. I added the `if:` check so it can be copied and used in other repos.

michael-hawker (approved):

Arlodotexe:
Needs #529

PR #548, opened 2024-06-03
Improved MarqueeText control

Avid29:
### Added Cycle Marquee Behavior option

When in cycle mode, an animation is played to transition between values when the `Text` property is changed.

https://github.com/CommunityToolkit/Labs-Windows/assets/9384045/e4a3e26a-1932-48dd-890d-63a5b8f092b8

### More predictable behavior

- Marquee animation no longer plays by default.
- Marquee animation is no longer started when the control's size changes.

### DIFF
--- a/components/MarqueeText/src/MarqueeText.cs
+++ b/components/MarqueeText/src/MarqueeText.cs
@@ -180,6 +180,7 @@ public partial class MarqueeText : Control
         // Update the animation to the stopped position.
         if (!_isActive)
         {
+            _marqueeStoryboard?.Stop();
             VisualStateManager.GoToState(this, MarqueeStoppedState, false);
 
             return false;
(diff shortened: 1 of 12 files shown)


components/MarqueeText/src/MarqueeText.cs
```diff
@@ -234,30 +267,29 @@ private bool UpdateAnimation(bool resume = true)
...
+        // Unbind events from the old storyboard and stop it before disposal.
         if (_marqueeStoryboard is not null)
         {
             _marqueeStoryboard.Completed -= StoryBoard_Completed;
+            _marqueeStoryboard?.Stop();
         }
 
         // Create new storyboard and animation
-        _marqueeStoryboard = CreateMarqueeStoryboardAnimation(start, end, duration, targetProperty);
+        _marqueeStoryboard = Behavior switch
```
Arlodotexe, line 284: We should consider the ways we can generalize this control so differences like this are handled in the style template instead of internally in the control. Might be a bit of work, but now that things are working under a variety of configurations it would be a good thing to do.

Arlodotexe (commented):
First pass review. As discussed in call, we'll do a bit of cleanup before merging this in. There are layout and extensibility complexities here that we'll want to resolve later, but the new cycling functionality works quite well as-is and should be made available for others.

Arlodotexe:
@michael-hawker Not here or now, but we should consider all the advantages we'd get building this on ItemsRepeater instead of manually repositioning two text segments via Storyboard/code-behind.

PR #554, opened 2024-06-27
Tooling Update

michael-hawker:
Looks like we haven't updated the tooling pointer in a while?

At least was out-of-sync on my machine.

Also had some minor tweaks to make to the TitleBar docs.

Though think we still don't display these great in the sample app, will take a look as I build it. See https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/70


michael-hawker:
Ugh, just noticed TitleBar is missing the Labs Icon and has the default X one...
image
Guess that's not part of the template?

image

Arlodotexe (approved):

michael-hawker:
Cool, we got a clean build, so I'll merge. Can open a PR for anything else like the TitleBar samples messing the TitleBar of the app issue I saw. Not sure if that's unique to trying to merge them together or will repro in the Labs all-samples gallery build too within just the single repo.

PR #556, opened 2024-07-01
Fix TitleBar Sample mucking up App's TitleBar

michael-hawker:
Add internal check for TitleBar code so that it only resets the App's TitleBar if it was the initializer of the setup originally

Before this change, if you opened the TitleBar sample the App's TitleBar customizations would be reset. After this change, that no longer occurs.


niels9001 (approved):

Arlodotexe:
Tested on WinAppSDK, all is well. Merging.

PR #576, opened 2024-10-11
Add support for modern .NET to AppServices library

Sergio0694:
This PR adds support for .NET 8 to the AppServices library (including [UWP on .NET 9](https://devblogs.microsoft.com/ifdef-windows/preview-uwp-support-for-dotnet-9-native-aot/)).

### DIFF
--- a/components/AppServices/src/CommunityToolkit.AppServices.targets
+++ b/components/AppServices/src/CommunityToolkit.AppServices.targets
@@ -1,5 +1,10 @@
 <Project>
 
+  <!-- Allow the source generators to detect whether an app on modern .NET is UWP or not -->
+  <ItemGroup>
+    <CompilerVisibleProperty Include="UseUwpTools" />
+  </ItemGroup>
+
   <!-- Get the analyzer from the CommunityToolkit.AppServices NuGet package -->
   <Target Name="CommunityToolkitAppServicesGatherAnalyzers">
     <ItemGroup>
--- a/components/AppServices/src/MultiTarget.props
+++ b/components/AppServices/src/MultiTarget.props
@@ -4,6 +4,6 @@
           MultiTarget is a custom property that indicates which target a project is designed to be built for / run on.
           Used to create project references, generate solution files, enable/disable TargetFrameworks, and build nuget packages.
         -->
-        <MultiTarget>uwp;netstandard;net8.0-windows10.0.17763.0</MultiTarget>
+        <MultiTarget>uwp;netstandard;</MultiTarget>
     </PropertyGroup>
 </Project>
--- a/components/AppServices/src/MultiTarget.props
+++ b/components/AppServices/src/MultiTarget.props
@@ -4,6 +4,6 @@
           MultiTarget is a custom property that indicates which target a project is designed to be built for / run on.
           Used to create project references, generate solution files, enable/disable TargetFrameworks, and build nuget packages.
         -->
-        <MultiTarget>uwp;netstandard;</MultiTarget>
+        <MultiTarget>uwp;netstandard;net8.0-windows10.0.17763.0</MultiTarget>
     </PropertyGroup>
 </Project>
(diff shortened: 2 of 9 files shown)


components/AppServices/src/MultiTarget.props
```diff
@@ -4,6 +4,6 @@
           MultiTarget is a custom property that indicates which target a project is designed to be built for / run on.
           Used to create project references, generate solution files, enable/disable TargetFrameworks, and build nuget packages.
         -->
-        <MultiTarget>uwp;netstandard;</MultiTarget>
+        <MultiTarget>uwp;netstandard;net8.0-windows10.0.17763.0</MultiTarget>
```
Sergio0694, line 7: @Arlodotexe not entirely sure about this, is this the right way to add a vanilla .NET 8 Windows TFM here?

Arlodotexe:
```suggestion
        <MultiTarget>uwp;netstandard;</MultiTarget>
```

Like the comment says, our custom `MultiTarget` property is used to enable and disable the TargetFrameworks we support per-platform.

The TFM for modern .NET support on UWP is with the other UWP TFMs in our Toolkit Tooling repo on the [`uwp-net8-windows`](https://github.com/CommunityToolkit/Tooling-Windows-Submodule/tree/uwp-net8-windows) branch. Majority of the changes here look fine, but you'll need that tooling branch.

components/AppServices/src/CommunityToolkit.AppServices.targets
```diff
@@ -1,5 +1,10 @@
 <Project>
 
+  <!-- Allow the source generators to detect whether an app on modern .NET is UWP or not -->
+  <ItemGroup>
+    <CompilerVisibleProperty Include="UseUwpTools" />
```
Arlodotexe, line 5: If this is needed for any modern .NET on UWP project that uses source generators, we should add this to our tooling as well. We have SampleGen source generators that'll need to work under modern .NET for UWP.

Arlodotexe (changes requested):
Broadly, we're still not quite ready to ship modern .NET for UWP in the Toolkit. AppServices is one component where WinUI 2 with modern .NET isn't needed, but TFM support is added via our tooling and applies to _all_ components in the solution. 

If you only need NuGet packages, you can build locally _and_ output NuGet packages with our component build script:
```pwsh
.\tooling\Build-Toolkit-Components.ps1 -MultiTargets uwp -winui 2 -Components AppServices -PreviewVersion local -NupkgOutput ./some/output/path -Release
``` 

While the code changes to AppServices here seem good, we need to make sure tooling is sorted for all components before we try to enable this _and_ merge it in.

components/AppServices/src/MultiTarget.props
Arlodotexe: Since the idea with this change is to add modern .NET support (TFM `net8.0-windows10*`) to AppServices wherever `uap` was used before (agnostic to uwp/wasdk/wpf), we'll continue to use the `MultiTarget` system for this.

We'll still have `HasWinUI` disabled, which downgrades the component to just the underlying TFM used, no WinUI. In the case of the `uwp` MultiTarget used for modern .NET, `UseUwp` and `UseUwpTools` [aren't enabled](https://github.com/CommunityToolkit/Tooling-Windows-Submodule/blob/uwp-net8-windows/MultiTarget/MultiTargetIdentifiers.props#L63) when `HasWinUI` is set to false.

components/AppServices/src/CommunityToolkit.AppServices.targets
Sergio0694: No, this is not general. We only need this here in these .targets, which are not evaluated by the project during build, but rather they're packaged in the NuGet and imported by _consuming_ projects. They're needed so the generator can see whether the project being built is UWP or not. But it's not needed otherwise. This should just remain here.

Arlodotexe (approved):

PR #587, opened 2024-11-06
Add an extension method for getting string attribute value from HtmlN…

euju-ms:
Adds an extension method for HtmlNode that gets the string attribute value and use it everywhere instead of using the underlying `HtmlNode.GetAttributeValue` method that's not AOT compatible.


nerocui:
LGTM

Arlodotexe (approved):

PR #586, opened 2024-11-06
Fixed possible NRE on MarkdownTextBlock.

azchohfi:
Fixes #585.

PR #588, opened 2024-11-11
Alzollin/markdown features

azchohfi:


Arlodotexe (commented):
Most of these changes seem pretty good. My only concern is that we're routing the event for handling clicks via the config object, which makes it more difficult to set from XAML.

azchohfi:
That is a fair point, but isn't that true for any property on the config? I did it this way since it wasn't that straightforward to add it to the markdowntextblock itself, since we have no easy reference to it.

Arlodotexe:
> That is a fair point, but isn't that true for any property on the config? I did it this way since it wasn't that straightforward to add it to the markdowntextblock itself, since we have no easy reference to it.

True, and anything on the config could be turned into a separate dependency property if we wanted binding support for the options.

The distinction seems to be that the config options (custom svg renderer, image provider) are a bit more advanced than a `LinkClicked` handler. These would likely need to be set up in code-behind anyway, whereas you might expect `LinkClicked` to show when declaring the type in XAML.

Arlodotexe:
This is also the behavior from the 7.x version of our `MarkdownTextBlock`.

azchohfi:
You are right. Done.

Arlodotexe (approved):
Great work! This will see plenty of usage and brings us closer to feature parity with the 7.x control. Let's ship it 🚀

nerocui:
Nice set of changes! Are we merging this soon?

michael-hawker:
@azchohfi would you have a chance to rebase this PR? Then we can merge:

image

I tried locally, but there are conflicting changes with your other PR #586 that I'm not 100% confident in resolving in MyHyperlinkButton.cs within the constructor.

michael-hawker:
Ah I see the conflict is from https://github.com/CommunityToolkit/Labs-Windows/pull/587 - that maybe helps as that is simple. I'll see if I can rebase this myself again.

michael-hawker:
Alright, all fixed! Set to auto-merge once the CI passes, thanks folks!

PR #597, opened 2024-11-14
MarkdownTextBlock: Allows passing an existing MarkdownDocument, closes #596

### LINKED ISSUE #596: MarkdownTextBlock: add alternative to Text property to pass in a MarkdownDocument to the control
Currently, the only way to set the content of a MarkdownTextBlock is via the `Text` property, which internally uses `Markdown.Parse` to convert to a `MarkdownDocument`. It would be nice to also allow users to pass in an existing `MarkdownDocument` as an alternative.
 
Example:

```xaml
<controls:MarkdownTextBlock
  MarkdownDocument="{x:Bind Doc, Mode=OneTime}" />
```

Benefits:
- Users can insert arbitrary logic after `Markdown.Parse(...)`, such as mutating the MarkdownDocument tree with decorations
- Users can provide a custom Markdown pipeline
- Users can preserve the MarkdownDocument for other operations, such as copying the content to HTML with `Markdown.ToHtml(...)`

jcoc611-microsoft:


nerocui:
I'd recommend thinking about exposing parsing method as well. We could allow the consumer to use their own copy of markdig for parsing, but it won't guarantee the same plugins in the pipeline which could result in parsing syntax that the renderer doesn't support.

PR #600, opened 2024-11-18
 Incremental Loading Behavior for ItemsView

Poker-sang:
Includes `NeedMoreItemTriggerBehavior` and `LoadMoreItemBehavior`

GIF2


Poker-sang:
I've tried to run
```pwsh
ApplyXamlStyling.ps1 -Main
```
And it said no file to style.

Why the workflow still fails

michael-hawker:
@Poker-sang it's this file:

> Checking: C:\a\Labs-Windows\Labs-Windows\components\Behaviors\samples\LoadMoreItemBehaviorSample.xaml
> FAIL

I'm not sure why the tool isn't fixing it locally for you, I wonder if we have an issue with the script when run against forks... hmm

What does running `git status` and `git diff origin/main your_branch_name_listed_from_git_status --name-only --diff-filter=ACM` provide back?

It could also maybe be an issue if folks don't have their remote labeled `origin` I guess too?

You should just be able to run `xstyler` directly (or ensure you have the extension in VS with the search to root option enabled):

```
dotnet tool run xstyler -c .\settings.xamlstyler -f components\Behaviors\samples\LoadMoreItemBehaviorSample.xaml
```

components/Behaviors/src/MultiTarget.props
```diff
@@ -0,0 +1,9 @@
+<Project>
+    <PropertyGroup>
+        <!--
+          MultiTarget is a custom property that indicates which target a project is designed to be built for / run on.
+          Used to create project references, generate solution files, enable/disable TargetFrameworks, and build nuget packages.
+        -->
+        <MultiTarget>uwp;wasdk;wpf;wasm;linuxgtk;macos;ios;android;</MultiTarget>
```
michael-hawker, line 7: Since this is only good for WinUI 3, remove `uwp` here. @Arlodotexe that should exclude it from making an entire UWP package for Uno too, eh? Then we wouldn't need the `#if WINAPPSDK` conditional, right?

Poker-sang: Yes this feature only works for wasdk. But sorry that I'm not familiar with `tooling` project, so don't know how exactly i should do. I hope you can help me fix this❤️thanks

Poker-sang:
> ```
> dotnet tool run xstyler -c .\settings.xamlstyler -f components\Behaviors\samples\LoadMoreItemBehaviorSample.xaml
> ```

Thanks, this works

components/Behaviors/src/MultiTarget.props
michael-hawker: @Arlodotexe maybe something we can lend a quick hand with this after the long weekend here?

Poker-sang:
Also revert `tooling` changes here

PR #603, opened 2024-11-20
MarkdownTextBlock: Enable/Disable built in extensions

jcoc611-microsoft:
First step to making MarkdownTextBlock extensions configurable. Makes built-in Markdig extensions that have been implemented in the renderer configurable via properties. They are all `false` by default, but I'm open to changing that.

Adds the following properties:

- UseAutoLinks
- UseEmphasisExtras
- UseListExtras
- UsePipeTables
- UseTaskLists

Example
```xaml
<controls:MarkdownTextBlock
  Grid.Row="3"
  Config="{x:Bind MarkdownConfig, Mode=OneTime}"
  Text="{x:Bind Text, Mode=OneTime}"
  UseAutoLinks="True"
  UseEmphasisExtras="True"
  UseListExtras="True"
  UsePipeTables="True"
  UseTaskLists="True" />
```

### DIFF
--- a/components/MarkdownTextBlock/samples/MarkdownTextBlockCustomSample.xaml
+++ b/components/MarkdownTextBlock/samples/MarkdownTextBlockCustomSample.xaml
@@ -1,13 +1,12 @@
 <!--  Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information.  -->
-<Page
-    x:Class="MarkdownTextBlockExperiment.Samples.MarkdownTextBlockCustomSample"
-    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
-    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-    xmlns:controls="using:CommunityToolkit.Labs.WinUI.MarkdownTextBlock"
-    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
-    xmlns:local="using:MarkdownTextBlockExperiment.Samples"
-    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
-    mc:Ignorable="d">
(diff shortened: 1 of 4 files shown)


Arlodotexe:
We'll want to add xmldoc comments to the new public-facing properties before merging.

Arlodotexe:
Looking through the available extensions and docs at https://github.com/xoofx/markdig/tree/master/src/Markdig.Tests/Specs, there's a few more we might want to include.

Namely:
- `HardlineBreak` will be useful for rendering LLM output.
- `AlertBlock` is needed for https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/70

nerocui:
For Mathematics support, I was just looking at that today. There's this wonderful library renders Latex, but it uses SkiaSharp which is not a dependency we want to take in the toolkit. Maybe we can leave off Latex support to third party or a separate package?

jcoc611-microsoft:
I added comments and added support for `HardlineBreak` since it works out of the box. For the other extensions, would probably address in separate PRs.

By the way, do we care about supporting changing the `UseFoo` properties during runtime? (i.e. should I bother to add a change callback to those properties and re-create the markdown pipeline & renderer when they change?)

Arlodotexe:
> For Mathematics support, I was just looking at that today. There's this wonderful library renders Latex, but it uses SkiaSharp which is not a dependency we want to take in the toolkit. Maybe we can leave off Latex support to third party or a separate package?
> 
> https://github.com/verybadcat/CSharpMath

Right, makes sense. Might be better to create a new Labs component to isolate the optional dependency, or to produce a third-party library instead. Let's leave this out while we weight the options.

Arlodotexe:
> By the way, do we care about supporting changing the `UseFoo` properties during runtime? (i.e. should I bother to add a change callback to those properties and re-create the markdown pipeline & renderer when they change?)

Good question. I don't think most users will expect this functionality until they need it. Since we're in Labs, let's save that for a future PR when it becomes needed.

Arlodotexe (approved):

Issue #547, opened 2024-06-02
MarkdownTextBlock MyCodeBlock renders empty RichTextBox as doesn't use StringBuilder for anything


JoeTomkinson:
### Describe the bug

The 'MyCodeBlock' in the pre-release of the MarkDownTextBlock doesnt actually use the lines of text for anything resulting in empty code blocks. I've attached the class snippet below so you can see what I mean.

components/MarkdownTextBlock/src/TextElements/MyCodeBlock.cs

```
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.
// See the LICENSE file in the project root for more information.

using Markdig.Syntax;

namespace CommunityToolkit.Labs.WinUI.MarkdownTextBlock.TextElements;

internal class MyCodeBlock : IAddChild
{
    private CodeBlock _codeBlock;
    private Paragraph _paragraph;
    private MarkdownConfig _config;

    public TextElement TextElement
    {
        get => _paragraph;
    }

    public MyCodeBlock(CodeBlock codeBlock, MarkdownConfig config)
    {
        _codeBlock = codeBlock;
        _config = config;
        _paragraph = new Paragraph();
        var container = new InlineUIContainer();
        var border = new Border();
        border.Background = (Brush)Application.Current.Resources["ExpanderHeaderBackground"];
        border.Padding = _config.Themes.Padding;
        border.Margin = _config.Themes.InternalMargin;
        border.CornerRadius = _config.Themes.CornerRadius;
        var richTextBlock = new RichTextBlock();

        if (codeBlock is FencedCodeBlock fencedCodeBlock)
        {
//#if !WINAPPSDK
//            var formatter = new ColorCode.RichTextBlockFormatter(Extensions.GetOneDarkProStyle());
//#else
//            var formatter = new ColorCode.RichTextBlockFormatter(Extensions.GetOneDarkProStyle());
//#endif
            var stringBuilder = new StringBuilder();

            // go through all the lines backwards and only add the lines to a stack if we have encountered the first non-empty line
            var lines = fencedCodeBlock.Lines.Lines;
            var stack = new Stack<string>();
            var encounteredFirstNonEmptyLine = false;
            if (lines != null)
            {
                for (var i = lines.Length - 1; i >= 0; i--)
                {
                    var line = lines[i];
                    if (String.IsNullOrWhiteSpace(line.ToString()) && !encounteredFirstNonEmptyLine)
                    {
[...]

PR #604, opened 2024-11-21
MarkdownTextBlock: Minor style improvements

jcoc611-microsoft:
Before:

image

After:

image

Changes:
- Lists have proper margins before and after
- List item bullets have proper padding and are right-aligned to the column
- Thematic breaks now have 100% width rather than hard-coded to 500px
- Headings now have proper margins before


jcoc611-microsoft:
Fixes some of the style issues mentioned in #547

Arlodotexe (approved):

PR #614, opened 2024-11-28
MarkdownTextBlock: Fix nested styles, sub & super

jcoc611-microsoft:
Markdown:
```md
# Headline with some^superscript^

**bold~sub~ and^super^**

regular~sub~ and^super^

*nested italics **and bold***
```

Before:
Screenshot 2024-11-28 131236

After:
Screenshot 2024-11-28 130857

Fixes:
- Nested styles (child styles incorrectly propagated to parent)
- Superscript and Subscript (used font variants, now use custom logic)
- Generic mechanism for children to inherit parent styles when necessary (e.g. heading containing sub/super which need to match the foreground and size)


nerocui (approved):

Arlodotexe (changes requested):
Tested locally on WinAppSDK, I'm not seeing the expected behavior from Superscript and Subscript syntax:

image


image

We should investigate why only the nested superscript rendered properly, so we can correc the issue.

PR #616, opened 2024-12-03
MarkdownTextBlock: Adds OnMarkdownParsed event

jcoc611-microsoft:
Adds an event `MarkdownTextBlock.OnMarkdownParsed`, which is useful because consumers cannot otherwise know when the MarkdownDocument has finished being generated from the markdown text input. It also technically allows consumers to mutate the markdown parsed tree *before* the render step takes place.

### DIFF
--- a/components/MarkdownTextBlock/src/MarkdownTextBlock.xaml.cs
+++ b/components/MarkdownTextBlock/src/MarkdownTextBlock.xaml.cs
@@ -55,8 +55,6 @@ public partial class MarkdownTextBlock : Control
         private set => SetValue(MarkdownDocumentProperty, value);
     }
 
-    public event EventHandler<LinkClickedEventArgs>? OnLinkClicked;
-
     internal void RaiseLinkClickedEvent(Uri uri) => OnLinkClicked?.Invoke(this, new LinkClickedEventArgs(uri));
 
     private static void OnConfigChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
@@ -116,6 +114,7 @@ public partial class MarkdownTextBlock : Control
             if (!string.IsNullOrEmpty(Text))
             {
                 this.MarkdownDocument = Markdown.Parse(Text, _pipeline);
+                OnMarkdownParsed?.Invoke(this, new MarkdownParsedEventArgs(this.MarkdownDocument));
                 _renderer.Render(this.MarkdownDocument);
             }
         }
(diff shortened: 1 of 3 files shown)


nerocui (approved):

components/MarkdownTextBlock/src/MarkdownTextBlock.xaml.cs
```diff
@@ -116,6 +114,7 @@ private void ApplyText(bool rerender)
             if (!string.IsNullOrEmpty(Text))
             {
                 this.MarkdownDocument = Markdown.Parse(Text, _pipeline);
+                OnMarkdownParsed?.Invoke(this, new MarkdownParsedEventArgs(this.MarkdownDocument));
```
azchohfi, line 117: I believe there is a 1-1 relationship between MarkdownTextBlock and its MarkdownDocument, so I would remove the Document property from the `MarkdownParsedEventArgs`, as you can get that from the sender.

jcoc611-microsoft: Every time you change the `Text` property of MarkdownTextBlock, technically you would get a brand new MarkdownDocument tree. If we also move the parsing to another thread, technically there would be no guarantees that the MarkdownDocument currently set on the control is the one corresponding to the given event, especially in streaming contexts

Arlodotexe: That's a good point @jcoc611-microsoft. While you could get the document via `sender.MarkdownDocument`, without a mutex or semaphore here the text *could* be updated concurrently and would create a separate `MarkdownDocument` instance for each call.

Since this could happen concurrently, there no guarantee that the `sender.MarkdownDocument` is the same one retrieved from `Markdown.Parse(Text, _pipeline);` on line 116 here.

Given this, I'd even advocate for storing the result of `Mardown.Parse` in its' own variable instead of `this.MarkdownDocument` when providing the value to the event.

Arlodotexe: > Given this, I'd even advocate for storing the result of `Mardown.Parse` in its' own variable instead of `this.MarkdownDocument` when providing the value to the event, though that's more of a nit.

Fixed in https://github.com/jcoc611-microsoft/Labs-Windows/commit/dc7bf037fba2d4f956b7abd4e338b4876d5abc42

Arlodotexe (approved):

PR #617, opened 2024-12-03
🧪 [Experiment] DependencyPropertyGenerator 

Arlodotexe:
### See #621

This PR introduces a new `DependencyPropertyGenerator` component that utilizes source generators on partial properties to generate `DependencyProperty` quickly in your projects. 

This went through some [community discussion](https://github.com/CommunityToolkit/Labs-Windows/discussions/449) before the current API design was finalized.  

Prerequisite PR: https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/234


components/DependencyPropertyGenerator/CommunityToolkit.DependencyPropertyGenerator.Tests/Test_DependencyPropertyGenerator.cs
```diff
@@ -0,0 +1,1962 @@
...
+                        ownerType: typeof(MyControl),
+                        typeMetadata: new global::Windows.UI.Xaml.PropertyMetadata(default(int)));
+
+                    /// <inheritdoc/>
+                    [global::System.CodeDom.Compiler.GeneratedCode("CommunityToolkit.WinUI.DependencyPropertyGenerator", <ASSEMBLY_VERSION>)]
+                    [global::System.Diagnostics.DebuggerNonUserCode]
+                    [global::System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
+                    public partial int Number
+                    {
+                        get => field;
```
michael-hawker, line 56: We should be calling `GetValue` eh and not using the `field` here? https://learn.microsoft.com/en-us/windows/uwp/xaml-platform/custom-dependency-properties#implementing-the-wrapper

We should also note that the methods aren't guaranteed to be called in the setter either.

michael-hawker:
It'd be helpful to have a summary of the features/syntax and such at least in the PR, if not in the doc or something. I had to find the test cases to see how it'd work. Glad to see the check for containing type of `DependencyObject` 🎉

Some general questions/thoughts:

- It'd be nice to have an analyzer/fixer for suggesting a property changed callback, so you don't have to remember the partial signature there to add one once you create a partial property + attribute.
- Should we have an analyzer warning if your property name ends in `Property` so you don't end up with a DP of `MyPropertyProperty`?
- Oh, I guess we can't name this `DependencyProperty` as it'd conflict with the platform type, eh? I was confused why `Generated` was there first, compared to our nice `ObservableProperty` we have in the MVVM Toolkit. Can't think of a better alternative right now though... 🤔
- We can add more helpers/analyzers from the docs to encode later, there's a lot of special cases/guidelines to shoot yourself with... e.g.
  - https://learn.microsoft.com/en-us/windows/uwp/xaml-platform/custom-dependency-properties#initializing-the-collection
  - setting an object as the default value, etc...
- Will the factory constructor pattern be supported for the default value? https://learn.microsoft.com/en-us/windows/uwp/xaml-platform/custom-dependency-properties#createdefaultvaluecallback
- Is Attached Property also planned still too?

components/DependencyPropertyGenerator/CommunityToolkit.DependencyPropertyGenerator.SourceGenerators/Constants/WellKnownTypeNames.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// <summary>
+    /// The fully qualified type name for the <c>XAML</c> namespace.
+    /// </summary>
+    public const string XamlNamespace =
+#if WINDOWS_UWP
+        "Windows.UI.Xaml";
+
+#else
+        "Microsoft.UI.Xaml";
+#endif
```
michael-hawker, line 31: @Sergio0694 would it be worth trying to see how this could work with WPF as well here? Theoretically it'd just be a different namescape here too, eh? `System.Windows` https://learn.microsoft.com/dotnet/api/system.windows.dependencyproperty?view=windowsdesktop-8.0 - or would we not be able to distinguish WinUI from WPF with the .NET TFM the same way unless we had a third package?

Sergio0694: Correct, yeah we'd need to ship a separate package just for WPF, also we have no tests, etc. 😅

michael-hawker: Yeah, that's what I thought and a larger discussion I've been thinking about for some other stuff too. Bubbled it up elsewhere to at least think about for 2025 as we move forward.

components/DependencyPropertyGenerator/CommunityToolkit.DependencyPropertyGenerator.SourceGenerators/Extensions/ITypeSymbolExtensions.cs
```diff
@@ -0,0 +1,139 @@
...
+            switch (symbol)
+            {
+                // Namespaces that are nested also append a leading '.'
+                case INamespaceSymbol { ContainingNamespace.IsGlobalNamespace: false }:
+                    BuildFrom(symbol.ContainingNamespace, in builder);
+                    builder.Add('.');
+                    builder.AddRange(symbol.MetadataName.AsSpan());
+                    break;
+
+                // Other namespaces (ie. the one right before global) skip the leading '.'
```
0x5bfa, line 109:
```suggestion
                // Other namespaces (i.e. the one right before global) skip the leading '.'
```

0x5bfa (commented):
Thank you for the hard work, left a few nit comment corrections..

michael-hawker:
@Sergio0694 was there anything left for your 'V1' of this to do, or should we merge the base in here to start testing it out in places to get some initial feedback?

Sergio0694:
Update: I added a code fixer too! 🎉

components/AppServices/src/CommunityToolkit.AppServices.csproj
```diff
@@ -8,12 +8,13 @@
     <PackageId>$(PackageIdPrefix).$(ToolkitComponentName)</PackageId>
     <HasWindowsUIXaml>false</HasWindowsUIXaml>
     <HasWinUI>false</HasWinUI>
+    <EnableXamlCompilerMismatchedTfmUnoWorkaround>false</EnableXamlCompilerMismatchedTfmUnoWorkaround>
 
     <!--
-      We also include 'net9.0-windows10.0.19041.0' explicitly for UWP as a workaround for https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/238.
-      We target 19041 and not 17763 to avoid an issue with NuGet resolving no assemblies due to the empty TFM lib folders for Uno platform (which should be removed).
+      We also include 'net9.0-windows10.0.17763.0' explicitly for UWP as a workaround for https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/238.
+      We target 17763 and not 17763 to avoid an issue with NuGet resolving no assemblies due to the empty TFM lib folders for Uno platform (which should be removed).
```
Arlodotexe, line 15: This is a bit confusing, does it need updated?

Sergio0694:
#### Moved to #624

PR #620, opened 2024-12-06
Add experiments list to readme

niels9001:
Add list of experiments to readme

(_Based on Discord discussion here: https://discord.com/channels/372137812037730304/671870147354427422/1314180640983486497_)


michael-hawker:
I can't remember which one is the 'main' one, probably best to have a separate specific Labs project, but it could be good to update and link to that sort of view as well. E.g. https://github.com/orgs/CommunityToolkit/projects/7/views/5

Not sure if those are embeddable... so we could just have it point to the board directly, but have a nice view in the readme (GH feature request???).

michael-hawker (approved):

PR #622, opened 2024-12-26
Trivia handling

Youssef1313:


Sergio0694 (commented):
Added some comments and fixed the merge conflict. I'm getting test failures though 🥲

image

Thoughts?

Youssef1313:
@Sergio0694 Sorry not home right now. I will check when I get back. I tested my change with one test only unfortunately.

I'd **guess** there may be some syntax editor call that should use the lambda overload to fix that error.

Youssef1313:
@Sergio0694 It should hopefully be better now

Sergio0694:
@Youssef1313 works great, thank you!! 😄

PR #647, opened 2025-02-27
Building only changed components.

azchohfi:


.github/workflows/build.yml
```diff
@@ -132,10 +133,28 @@ jobs:
...
+          }
+          else {
+            echo "CHANGED_COMPONENTS_LIST='all'" >> $env:GITHUB_ENV
+          }
+  
       # Generate full solution with all projects (sample gallery heads, components, tests)
       - name: Generate solution with ${{ matrix.multitarget }} gallery, components and tests
         working-directory: ./
-        run: powershell -version 5.1 -command "./tooling/GenerateAllSolution.ps1 -MultiTargets ${{ matrix.multitarget }} ${{ env.ENABLE_DIAGNOSTICS == 'true' && ' -UseDiagnostics' || '' }} -WinUIMajorVersion ${{ matrix.winui }}" -ErrorAction Stop
+        run: powershell -version 5.1 -command "./tooling/GenerateAllSolution.ps1 -MultiTargets ${{ matrix.multitarget }} ${{ env.ENABLE_DIAGNOSTICS == 'true' && ' -UseDiagnostics' || '' }} -Components ${{ env.CHANGED_COMPONENTS_LIST }} -WinUIMajorVersion ${{ matrix.winui }}" -ErrorAction Stop
```
michael-hawker, line 157: @Arlodotexe how does the input to the components parameter here in our script work with dependencies? i.e. if the only change is in X but that depends on component Y (in labs I guess), then is slngen/msbuild smart enough to know that Y is built too? I guess that'd mean we'd end up with packages for X and Y if only X changed. But that's not really a common thing we do here (compared to the main repo).

And I think this is just a process we'd use here in labs, so we could push things when changed to NuGet.org vs. in the main toolkit where we can just snap a preview to NuGet anyway and have nightlies in the custom feed if folks still need those, eh?

Arlodotexe: > if the only change is in X but that depends on component Y (in labs I guess), then is slngen/msbuild smart enough to know that Y is built too?

Yes, you can specify a single component that references other components (such as SettingsControls in mainline WCT) and the dependencies (Extensions, Converters, etc) will get picked up. 

> I guess that'd mean we'd end up with packages for X and Y if only X changed. 

No, the package build script `./toolkit/Build-Toolkit-Components.ps1` doesn't use slngen in any capacity. It will only build the components you supply in the list, including nuget package and binlog output.

michael-hawker:
@Arlodotexe any concerns if we just merge this and test it out with other changes that happen to see how it works in practice?

Arlodotexe:
> @Arlodotexe any concerns if we just merge this and test it out with other changes that happen to see how it works in practice?

It appears to be running without issue in the CI here, but since it's fully untested I'm wary of merging into `main`. 

There're two roads to take here:
- Move the code into a script file that can be reused and tested
- Merge this and watch it like a hawk in case it breaks

michael-hawker:
@Arlodotexe I approved your new tooling 281 the other week, can that be merged so we can get this updated and merged in to move forward on it? Thanks!

michael-hawker (approved):

PR #674, opened 2025-05-09
Add `DisableHtml` property to MarkdownTextBlock

michael-hawker:
Calls DisableHtml method on Markdig pipeline, as well as not adding any Html based renderer, HtmlBlockRenderer, HtmlEntityInlineRenderer, and HtmlInlineRenderer. When enabled, results in any html tags being ignored and plain-text parts remaining, unformatted.

Committing Patch on behalf of Ameya Kulkarni

### DIFF
--- a/components/MarkdownTextBlock/samples/MarkdownTextBlockCustomSample.xaml
+++ b/components/MarkdownTextBlock/samples/MarkdownTextBlockCustomSample.xaml
@@ -1,4 +1,4 @@
-<!--  Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information.  -->
+﻿<!--  Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information.  -->
 <Page x:Class="MarkdownTextBlockExperiment.Samples.MarkdownTextBlockCustomSample"
       xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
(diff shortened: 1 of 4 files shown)


michael-hawker:
We should add a tracking issue to have the Markdown sample expose checkbox options for each of these settings properties so folks can see the behavior change in the output. (Though I think we'd have to make the control re-create the pipeline builder, I think they're all assumed to be set on initialization? Not 100% sure about that.) May be easier if we can switch to using the DependencyProperyGenerator later for all this stuff. :)

ameyak23 (approved):

michael-hawker:
FYI @Arlodotexe I think there's an issue with the scoped component build here from #647 (and related). Looks like a part of the CI is still trying to build all the components, but then they're not loaded of course... At least that's what appears to be the case at first glance.

Arlodotexe:
> FYI @Arlodotexe I think there's an issue with the scoped component build here from #647 (and related). Looks like a part of the CI is still trying to build all the components, but then they're not loaded of course... At least that's what appears to be the case at first glance.

It looks like it's tripping on the Test project. The changes in #647 enabled CI to include only components that were updated in the generated solution, but the generated solution also includes a hardcoded wildcard reference to all test shproj in all components.

Since running tests is an integral part of the CI, we cannot simply forgoe building tests to fix the issue. We'll need to fix solution generation to selectively include the tests for the requested components.

Arlodotexe (approved):

michael-hawker:
Thanks Arlo! Some more work may be needed to support #675 (as I don't know if the control fully listens to all those properties changing dynamically), we can test now at least! :)

PR #684, opened 2025-06-10
MarkdownTextBlock: Rewrite lists to be text-based

jcoc611-microsoft:
Before, Lists were implemented using `InlineUIContainer` and grid layouts. This looks great, but breaks selection and accessibility. This PR rewrites lists all-up to use a combination of Paragraph, the list bullet character (either a "dot" or the index number/alpha/roman), and smart use of margins and indents.

`WinUIRenderer` keeps track of lists in a stack fashion, and `MyParagraph` uses the info from the renderer to offset the text and prepend the bullet character(s).


components/MarkdownTextBlock/src/Renderers/WinUIRenderer.cs
```diff
@@ -25,6 +27,10 @@ public MarkdownConfig Config
     }
     public MarkdownTextBlock MarkdownTextBlock { get; }
 
+    internal bool IsFirstBulletItem { get; set; }
```
Arlodotexe: The assignment of `IsFirstBulletItem` and `IsLastBulletItem` from outside the class raises a few questions about the relationship between `WinUIRenderer` and `ListRenderer` and how/where data should flow between them.

It doesn't appear that these are used outside of the `ListRenderer`, so I have two possible suggestions here:
- We could make these local variables in the `ListRenderer.Write` method and remove these props from `WinUIRenderer`.
- If we have reason to keep these variables as internal properties on `WinUIRenderer`, we should add xmldoc with a summary and a few remarks containing the rationale, since it's not immediately clear.

michael-hawker (changes requested):
I love the idea of translating lists into text instead of custom UI, makes a lot of sense and resolves a lot of issues with accessibility and copying.

Seeing how this works, I think the crux of the issue is that a List is contained in a Paragraph and we need to modify the properties of the parent paragraph to make this work/look good.

I'm wondering if we can be better here about using parsing to detect this or having the List Renderer make/be a paragraph vs. having to connect state between the paragraph and lists between the renderer. @nerocui any insights here?

I'd have to look at the parsing tree here (I wrote a tool for that actually a few weekends back, but still need to publish it.) Like I just wonder if in all cases a list is just encapsulated by a paragraph and any text before/after is a separate paragraph. If so, we should just make a dedicated "BulletedParagraph" type which will just output the final result that he Listrenderer does vs. it being a thing which needs to coordinate with the general paragraph.

components/MarkdownTextBlock/src/Renderers/WinUIRenderer.cs
jcoc611-microsoft: oops these are not needed anymore, removed

Arlodotexe (approved):
Tested locally, no bugs or regressions. Text selection across and between lists now works as expected.

Arlodotexe:
New issue opened to cleanup the state management. Should give us space to continue the discussion properly and allow us to close this off.

michael-hawker (approved):
As discussed, we'll merge for now and work towards to see if we can clean-up the state a bit in the future. Functionality wise, worked great.

PR #688, opened 2025-06-30
Cleanup samples for MarkdownTextBlock

Arlodotexe:
This PR makes changes to the MarkdownTextBlock sample/doc:
- Splits the CustomMarkdownTextBlockSample into separate samples, one for feature showcase and another for the live editor.
- Cleans up templated text from the markdown doc
- Acts as the first non-smoke test of the incremental build fixes made in https://github.com/CommunityToolkit/Labs-Windows/pull/687

### DIFF
--- a/components/MarkdownTextBlock/samples/MarkdownTextBlockLiveEditorSample.xaml.cs
+++ b/components/MarkdownTextBlock/samples/MarkdownTextBlockLiveEditorSample.xaml.cs
@@ -32,7 +32,21 @@ public sealed partial class MarkdownTextBlockLiveEditorSample : Page
     {
         this.InitializeComponent();
         _config = new MarkdownConfig();
-        MarkdownTextBox.Text = "# Hello World\n\nTry typing some **markdown** here!\n\n- Item 1\n- Item 2\n- [ ] Task item\n\n> This is a blockquote\n\n```csharp\nConsole.WriteLine(\"Hello, World!\");\n```";
+        MarkdownTextBox.Text = """
+            # Hello World
+
+            Try typing some **markdown** here!
+
+            - Item 1
+            - Item 2
+            - [ ] Task item
+
+            > This is a blockquote
+
+            ```csharp
+            Console.WriteLine("Hello, World!");
+            ```
+            """;
         MarkdownTextBlock.OnLinkClicked += MarkdownTextBlock_OnLinkClicked;
     }
 


Arlodotexe (commented):
In the future, we might want to consider ways to exclude the "live editor" sample from the published MSLearn concept docs if we don't have a way to render the Uno WASM component inline on MSLearn.

components/MarkdownTextBlock/samples/MarkdownTextBlockLiveEditorSample.xaml.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        get => _config;
+        set => _config = value;
+    }
+
+    public MarkdownTextBlockLiveEditorSample()
+    {
+        this.InitializeComponent();
+        _config = new MarkdownConfig();
+        MarkdownTextBox.Text = "# Hello World\n\nTry typing some **markdown** here!\n\n- Item 1\n- Item 2\n- [ ] Task item\n\n> This is a blockquote\n\n```csharp\nConsole.WriteLine(\"Hello, World!\");\n```";
```
michael-hawker: Does a block quote work here for embedded newlines to make this a bit easier to read? i.e.

```cs
Text = """
    # Hello World
    Try typing...
    """
```

components/MarkdownTextBlock/samples/MarkdownTextBlockLiveEditorSample.xaml
```diff
@@ -11,23 +11,22 @@
...
-            <RowDefinition Height="auto" />
+            <RowDefinition Height="*" />
         </Grid.RowDefinitions>
         <TextBlock Grid.Row="0"
                    Margin="0,0,0,12"
                    FontSize="16"
                    FontWeight="Bold"
-                   Text="Try it live!" />
+                   Text="Try it live! Type in the text box below and see the markdown rendered in real-time:" />
         <Grid Grid.Row="1">
```
michael-hawker, line 21: Should we collapse this grid down? Would it just be easier to have the textbox on top and then the live preview on the bottom to better make use of the vertical space? Or maybe put the textbox on the left and the markdown preview on the right more like a markdown editor?

components/MarkdownTextBlock/samples/MarkdownTextBlockLiveEditorSample.xaml.cs
Arlodotexe: Resolved in 21e53624f341a0a9b90a6063b627b326c945caa8

components/MarkdownTextBlock/samples/MarkdownTextBlock.md
```diff
@@ -25,10 +25,12 @@ icon: Assets/MarkdownTextBlock.png
...
-The Toolkit is built with templated controls. This provides developers a flexible way to restyle components
-easily while still inheriting the general functionality a control provides. The examples below show
-how a component can use a default style and then get overridden by the end developer.
+> [!Sample MarkdownTextBlockExampleSample]
 
-> [!Sample MarkdownTextBlockCustomSample]
+## Interactive Live Editor
+Try typing markdown and see it rendered in real-time:
+
+> [!Sample MarkdownTextBlockLiveEditorSample]
```
michael-hawker, line 36: Should the interactive editor go above the other so it's easier for folks to try their own markdown without having to scroll to the bottom?

michael-hawker (commented):
Couple of small comments

components/MarkdownTextBlock/samples/MarkdownTextBlock.md
Arlodotexe: This was what I tried first, but since markdown-inlined samples have a max control height, the long nested scrollviewer on bottom created a lot of visual noise when scrolling to and away from it, especially when trying to reference the feature showcase while making markdown edits. This arrangement mitigates that by having a bottom sample that doesn't scroll internally.

components/MarkdownTextBlock/samples/MarkdownTextBlockLiveEditorSample.xaml
Arlodotexe: Vertical space is already pretty constrained when viewing the sample inside the markdown document, since they have a set max height. This is a good suggestion for the full-window view but would impair usability on the markdown doc view. We could consider an adaptive control trigger instead.

components/MarkdownTextBlock/samples/MarkdownTextBlock.md
Arlodotexe: Since we're rendering Markdown examples within a Markdown document here, I think the ideal setup here might be to remove the MaxHeight constraint on the feature showcase sample so we can move it to the bottom while avoiding the nested scrollviewer noise.

michael-hawker: I guess we'd need a flag for that as a sample option? (that'd touch the submodule and make this harder as a litmus test).

Maybe just leave it for now and we can come back to it later?

michael-hawker (approved):

PR #689, opened 2025-07-01
Run incremental component check earlier, save additional CI time

Arlodotexe:
Updates build.yml to run incremental component check as early as possible-- before uno-check, dotnet restore, msbuild setup, sdk installation, and machine configuration (pagefile, usermode dumps).

Since our incremental component checks use pwsh instead of msbuild, we can save an addition 4-6 minutes of CI time per multitarget by skipping these dependency install steps.


.github/workflows/build.yml
```diff
@@ -93,6 +108,7 @@ jobs:
           Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps' -Name 'DumpType' -Type DWord -Value '2'
 
       - name: Install .NET SDK v${{ env.DOTNET_VERSION }}
+        if: ${{ env.ENABLE_DIAGNOSTICS == 'true' }}
```
michael-hawker: Don't we always need to install the .NET SDK?

Arlodotexe: We don't need the .NET SDK to do an incremental component check, all the work is done by git and powershell.

Highlighted from OP:
> Since our incremental component checks use pwsh instead of msbuild

Arlodotexe:
> Is there a case where the buildable components is empty?

It didn't get far enough to do the component check. The error is coming from git:
```
Run $changedComponents = $(./tooling/Get-Changed-Components.ps1 8c321638c2ead1e02181fc460f9e71d7ead9d946 fd2ad1d6b79bff39cb8f3ea5311684b2b6d09dca)
From https://github.com/CommunityToolkit/Labs-Windows
 * branch              main       -> FETCH_HEAD
fatal: Invalid symmetric difference expression 8c321638c2ead1e02181fc460f9e71d7ead9d946...fd2ad1d6b79bff39cb8f3ea5311684b2b6d09dca
fatal: Invalid symmetric difference expression 8c321638c2ead1e02181fc460f9e71d7ead9d946...fd2ad1d6b79bff39cb8f3ea5311684b2b6d09dca
```

Looking at these 8c321638c2ead1e02181fc460f9e71d7ead9d946 and fd2ad1d6b79bff39cb8f3ea5311684b2b6d09dca, these commits are identical both in content and in the parent commit. The only discernable difference is that the latter is a commit that resides on our current `ci/incremental-builds/improvement/skip-uno-check` branch. 

In a normal push, the before commit is a direct ancestor of the after commit. However, with a force push, you're rewriting branch history.

Since the before commit is often no longer part of the branch's history, and it has no direct relationship with the after commit and cannot be compared symmetrically. This is the error we're seeing here. 

That makes me wonder whether rebases are going to work as expected. I'll need to investigate before we can settle on a solution.

Arlodotexe:
@michael-hawker It looks like that fixed the issue with force pushes! We should be able to merge once CI passes 😄

michael-hawker (approved):

PR #690, opened 2025-07-15
Add weekly merge workflow and update CI triggers

Arlodotexe:
This PR adds a `weekly-merge.yml` workflow file that merges `main` into `rel/weekly`. 

When this merge happens, our usual `build.yml` will be triggered and should diff with the merge from the week prior (the last push).

If no incremental changes are found by our `build.yml`, the build will exit early as usual.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,13 +8,19 @@ name: CI
 on:
   # Triggers the workflow on push or pull request events but only for the main branch
   push:
-    branches: [ main ]
+    branches: [ main, 'rel/weekly' ]
   pull_request:
     branches: [ main ]
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
   merge_group:
+
+  # Allow this workflow to be triggered by scheduled release workflows
+  workflow_run:
+    workflows: [ WeeklyMerge ]
+    types:
+      - completed
 
 env:
   DOTNET_VERSION: ${{ '9.0.x' }}
--- a/.github/workflows/weekly-merge.yml
+++ b/.github/workflows/weekly-merge.yml
@@ -1,0 +1,36 @@
+name: WeeklyMerge
+
+on:
+  schedule:
+    # Runs every Wednesday at 08:00 UTC (midnight PST / 1:00 AM PDT)
+    - cron: '0 8 * * 3'
+
+  # Allows manual triggering for convenience
+  workflow_dispatch:
+
+jobs:
+  weekly-release:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          # Use a token with write permissions to push to the branch
+          token: ${{ secrets.GITHUB_TOKEN }}
+          fetch-depth: 0 # Fetch all history for merging
+
+      - name: Configure Git
+        run: |
+          git config user.name 'github-actions[bot]'
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
+
+      - name: Merge main into rel/weekly
+        run: |
+          git fetch origin
+          git checkout rel/weekly
+          git reset --hard origin/rel/weekly
+          git merge --no-ff origin/main -m "Weekly merge of main into rel/weekly"
+          git push origin rel/weekly


.github/workflows/build.yml
```diff
@@ -8,14 +8,20 @@ name: CI
...
   pull_request:
     branches: [ main ]
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
   merge_group:
 
+  # Allow this workflow to be triggered by scheduled release workflows
+  workflow_run:
+    workflows: [ WeeklyMerge ]
```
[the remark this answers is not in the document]
Arlodotexe: The existing code already 1:1 matches the name field defined inside of `weekly-merge.yml` If the workflow name changes, I expect us to find all places where the previous workflow name was used so they can be updated.

.github/workflows/weekly-merge.yml
```diff
@@ -0,0 +1,36 @@
...
+          # Use a token with write permissions to push to the branch
+          token: ${{ secrets.GITHUB_TOKEN }}
+          fetch-depth: 0 # Fetch all history for merging
+
+      - name: Configure Git
+        run: |
+          git config user.name 'github-actions[bot]'
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
+
+      - name: Merge main into rel/weekly
```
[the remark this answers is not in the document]
Arlodotexe: Counter-arguments here:

1. **Simple merge scenario** - We're merging main into rel/weekly weekly. If there are conflicts, it means someone committed directly to rel/weekly (which shouldn't happen in this workflow)
2. **Fail-fast is good** - If there are conflicts, we want the workflow to fail loudly so someone investigates
3. **Easy to fix manually** - If it fails, we can just run the workflow manually (workflow_dispatch) after resolving conflicts
4. **Adds complexity** - Error handling would make the workflow much more complex for a rare edge case

Given this, let's go without.

.github/workflows/build.yml
michael-hawker, line 1: On L81 we'll need to adjust the version property to also check if it's a rel build. I think we'll use the same format?

michael-hawker, line 1: We also need to copy over the signing and push to NuGet bits from the main repo L313+, though here we'll only sign on the rel branches.

michael-hawker (changes requested):
Think we need a couple of more bits.

.github/workflows/build.yml
Arlodotexe: Good catch! I'll do a once-through and make sure any reference to main are updated to include `rel/*`.

Arlodotexe: Fixed in f267328d8329d7a75d9253ab17015b8d789d2638

Arlodotexe: I see, so we're going straight to the public NuGet package in the weekly release, whereas `main` will continue to be pushed to the Labs feed on-demand. Let's add that.

Arlodotexe: Added in 2780c5c41a36667500773db13eadb62a8518603b

.github/workflows/build.yml
```diff
@@ -296,7 +302,7 @@ jobs:
           dotnet nuget push "*.nupkg" --api-key dummy --source PullRequests --skip-duplicate
           
       - name: Push packages (main)
-        if: ${{ github.ref == 'refs/heads/main' && env.HAS_BUILDABLE_COMPONENTS == 'true' }}
+        if: ${{ (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/rel/')) && env.HAS_BUILDABLE_COMPONENTS == 'true' }}
         run: |
           dotnet nuget add source https://pkgs.dev.azure.com/dotnet/CommunityToolkit/_packaging/CommunityToolkit-Labs/nuget/v3/index.json `
```
michael-hawker: I'm thinking for the rel branch we push to MainLatest, helps us keep track of the signed vs. unsigned packages easier for validation before we push to NuGet and allows more visibility to anyone using the MainLatest feed as a central place to collect all most recent toolkit packages?

.github/workflows/build.yml
```diff
@@ -320,6 +326,103 @@ jobs:
           name: build-logs-winui${{ matrix.winui }}
           path: ./*.*log
 
+  sign:
+    needs: [package]
+    if: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/rel/') }}
```
michael-hawker: For Labs we should only sign the release packages we'll make once a week, for everything else they can remain as-is, don't need to take that extra time.

.github/workflows/build.yml
```diff
@@ -320,6 +326,103 @@ jobs:
...
+      - name: Sign Packages
+        run: >
+          ./tools/sign code azure-key-vault
+          **/*.nupkg
+          --base-directory "${{ github.workspace }}/packages"
+          --file-list "${{ github.workspace }}/SignClientFileList.txt"
+          --timestamp-url "http://timestamp.digicert.com"
+          --publisher-name ".NET Foundation"
+          --description "Windows Community Toolkit"
+          --description-url "https://github.com/CommunityToolkit/Windows"
```
michael-hawker:
```suggestion
          --description "Windows Community Toolkit Labs"
          --description-url "https://github.com/CommunityToolkit/Labs-Windows"
```

.github/workflows/build.yml
michael-hawker: Actually, never mind they'll get pushed to the main latest feed in the signing step below, so I think we should skip the push here for rel and leave it as it was.

.github/workflows/build.yml
Arlodotexe: Fixed in 58ac7474833886e090d07f04584fedfb917e1776

.github/workflows/build.yml
Arlodotexe: Fixed in https://github.com/CommunityToolkit/Labs-Windows/pull/690/commits/c814f7334f115e3face56a34a9f2cb4ef557281e

michael-hawker (approved):
LGTM

PR #696, opened 2025-07-29
Use PRs, fix inherited ref context for scheduled releases

Arlodotexe:
Switches our weekly releases from direct merges to PR-based merges to fix cron job context inheritance issues.

In the setup originally introduced in #690:
- The cron job always runs under `main` by default
- The pushes made to `rel/weekly` triggered the build.yml workflow, but it also inherited the original `main` ref as context when we wanted it to be `rel/weekly`. 
- Creating a PR from `main` into `rel/weekly` should ensure that the workflow is run on the base `rel/weekly` ref once merged.
- Since we're no longer directly doing a merge in the cron CI, we've renamed the workflow to `scheduled-releases` instead of `weekly-merge`. The job in this workflow remains named `weekly-release`, with the rename leaving room for scheduled releases other than weekly.

### Added details from planning:
According to [these official docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-github-cli), workflows can use the `gh` cli out of the box.
- Inline values are likely enough here.
- We'll want to set the head to `main` and the base to `rel/weekly`
- The title should include the date and "Scheduled weekly release"
- The body doesn't need to include much, only the bare essentials for identifying the PR from the workflow.
- We should auto-assign someone with release gate permissions to be the reviewer
- We can use `--fill-verbose` to also populate the body with the commit messages that are being released.
- We'll want to set auto-merge for this, which we can do via `gh pr merge <pr#> --auto --rebase`


Arlodotexe:
> Thanks Arlo, so this should create a PR that is set to auto-merge. So assuming no issues, we should see:
> 
> 1. This new chron kick-off
> 2. A new PR created
> 3. That PR should get built by the main build script under the context of the rel branch (but as a PR - we should double-check the logic there)
> 4. If all succeeds, it should auto-merge
> 5. An integration build fires in the rel context still again
> 6. Then await us on for the publish gate to nuget, as we do with any release.

Yes, exactly this with one missing detail-- someone with write permission will need to approve the PR.

.github/workflows/scheduled-releases.yml
```diff
@@ -0,0 +1,51 @@
...
+        shell: pwsh
+        run: |
+          # Create PR from main to rel/weekly, capture created URL
+          $PR_URL = $(gh pr create `
+            --base rel/weekly `
+            --head main `
+            --title "Scheduled weekly release $(Get-Date -Format 'yyyy-MM-dd')" `
+            --body "Automated weekly release PR created by scheduled workflow." `
+            --fill-verbose `
+            --reviewer michael-hawker)
```
michael-hawker, line 38: Can we use an existing team here like `ToolkitTeam`? Or leave it blank?

I guess that's the catch, is we need someone to approve this PR, eh?

michael-hawker (approved):

.github/workflows/scheduled-releases.yml
Arlodotexe: That's correct. I coded it to request you since you normally need to hit the nuget gate approval anyway, but anyone with write perms can approve as-is and we could tweak branch protection to accommodate this any time.

PR #705, opened 2025-08-04
Use reusable workflow call instead of PRs for incremental builds

Arlodotexe:
This PR is a follow-up to https://github.com/CommunityToolkit/Labs-Windows/pull/696 which:
- Reverts to git-based merges instead of PRs
- Adds a reusable build workflow trigger that directly specifies the `ref` that `build.yml` should run on.

This is supported according to the docs at https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#calling-a-reusable-workflow


michael-hawker (approved):
🤞

PR #715, opened 2025-08-20
added support foer base64 (dataurl) and file:/// sources url for markdown

Fefedu973:


Fefedu973:
@dotnet-policy-service agree

PR #716, opened 2025-08-21
Adding overridable MinWidth for TitleBarContent

niels9001:
There was no way for an app to override the TitleBarContentMinWidth - this is now configurable by using lightweight styling.

### DIFF
--- a/components/TitleBar/src/TitleBar.xaml
+++ b/components/TitleBar/src/TitleBar.xaml
@@ -7,7 +7,7 @@
 
     <x:Double x:Key="TitleBarCompactHeight">32</x:Double>
     <x:Double x:Key="TitleBarTallHeight">48</x:Double>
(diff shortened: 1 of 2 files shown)


components/TitleBar/src/TitleBar.xaml
```diff
@@ -7,7 +7,7 @@
 
     <x:Double x:Key="TitleBarCompactHeight">32</x:Double>
     <x:Double x:Key="TitleBarTallHeight">48</x:Double>
-    <x:Double x:Key="TitleBarContentMaxWidth">360</x:Double>
+    <x:Double x:Key="TitleBarContentMinWidth">360</x:Double>
```
michael-hawker, line 10: Funny that this is here but we weren't using this.

michael-hawker (approved):
LGTM

Issue #661, opened 2025-03-20
Inconsistent Package Identities


michael-hawker:
### Describe the bug

The following are missing .Controls. in their Package Ids:

- OpacityMaskView
- Ribbon
- TitleBar

i.e. they're missing this
https://github.com/CommunityToolkit/Labs-Windows/blob/8d0dc1c348833a064c9edfd01aa8d221042392c8/components/DataTable/src/CommunityToolkit.WinUI.Controls.DataTable.csproj#L16

This breaks the links to the packages in ADO from the Sample App with https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/275

There are other larger inconsistencies here we should investigate fixing for 8.3.

Ideally, all projects and namespaces are simple and aligned to the practices we have in the main repo (no Labs). Only thing that has Labs is the PackageId for when pushed to NuGet feeds.

### Steps to reproduce

```text
-
```

### Expected behavior

Sample App can link to these packages consistently based on project name pattern it has access to.

### Code Platform

- [ ] UWP
- [ ] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [ ] Windows 11 21H2 (Build 22000)
- [ ] Other (specify)

### App minimum and target SDK version

- [ ] Windows 10, version 1809 (Build 17763)
- [ ] Windows 10, version 1903 (Build 18362)
- [ ] Windows 10, version 1909 (Build 18363)
- [ ] Windows 10, version 2004 (Build 19041)
- [ ] Other (specify)

### Help us help you

Yes, I'd like to be assigned to work on this item.

PR #719, opened 2025-08-26
Bumping version number and adding packageid

niels9001:
Closes: #661

### DIFF
--- a/components/TitleBar/src/CommunityToolkit.WinUI.Controls.TitleBar.csproj
+++ b/components/TitleBar/src/CommunityToolkit.WinUI.Controls.TitleBar.csproj
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <ToolkitComponentName>TitleBar</ToolkitComponentName>
     <Description>This package contains TitleBar.</Description>
-    <Version>0.0.1</Version>
+    <Version>0.0.2</Version>
 
     <!-- Rns suffix is required for namespaces shared across projects. See https://github.com/CommunityToolkit/Labs-Windows/issues/152 -->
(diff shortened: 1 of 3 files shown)


michael-hawker:
Thanks @niels9001! 

FYI @Arlodotexe looks like CI is failing due to GetChangedComponents (multiple here)

Did we want to update the namespaces too? e.g.

https://github.com/CommunityToolkit/Labs-Windows/blob/00d77f2d6f8e91746968022c0441908a63213d49/components/MarkdownTextBlock/src/MarkdownTextBlock.xaml.cs#L13

Or open another ticket to track that and update separately?

components/TitleBar/src/CommunityToolkit.WinUI.Controls.TitleBar.csproj
```diff
@@ -2,12 +2,16 @@
   <PropertyGroup>
     <ToolkitComponentName>TitleBar</ToolkitComponentName>
     <Description>This package contains TitleBar.</Description>
-    <Version>0.0.1</Version>
+    <Version>0.0.2</Version>
```
michael-hawker, line 5: @Arlodotexe do we still need version here? This should just be handled automatically now, eh?

Arlodotexe: @michael-hawker We have automated versioning via https://github.com/CommunityToolkit/Tooling-Windows-Submodule/blob/786cf7bea168a37dbdeea2cf9acabf9d1cd35793/ToolkitComponent.SourceProject.props#L16, if it's set manually like this then it overrides the default computed version value.

Tooling supports both, it depends on our preference for this particular package.

PR #726, opened 2025-08-29
Use tag-based trigger for scheduled release build

Arlodotexe:
Refactor scheduled release build to trigger based on tags pushed in cron workflow instead of a reusable workflow.


michael-hawker (approved):
Signed off but not working in CI, some typo maybe?

PR #733, opened 2025-09-23
TitleBar flow direction handling

Avid29:
Added logic to automatically adjust the extended WindowStyles to match the TitleBar flow direction.

image


Arlodotexe (changes requested):
@Avid29 Tested WADSK locally and it appears that in RTL the opened new window can only be closed by the bottom right (bottom half?) of the red X close button. The top half of the close button doesn't work. 

We should see whether this is an easy fix before closing out this PR.

PR #734, opened 2025-09-25
ColorAnalyzer Component Creation

Avid29:
Adds the first iteration of the AccentAnalyzer, a helper which allows for binding to colors extracted from images or other UIElements.

Example:
image


Avid29:
Experiment Issue created
#736

Avid29:
@Arlodotexe Makes complete sense to me

Arlodotexe:
This is very strange, we're seeing issues with `Marquee` here stemming from https://github.com/CommunityToolkit/Labs-Windows/pull/732:
```
         C:\a\Labs-Windows\Labs-Windows\components\Marquee\src\Marquee.Properties.cs(5,18): error CS0234: The type or namespace name 'Text' does not exist in the namespace 'Windows.UI' (are you missing an assembly reference?) [C:\a\Labs-Windows\Labs-Windows\components\Marquee\src\CommunityToolkit.WinUI.Controls.Marquee.csproj::TargetFramework=uap10.0.17763]
         C:\a\Labs-Windows\Labs-Windows\tooling\GlobalUsings_WinUI.cs(11,33): error CS0234: The type or namespace name 'Collections' does not exist in the namespace 'Windows.Foundation' (are you missing an assembly reference?) [C:\a\Labs-Windows\Labs-Windows\components\Marquee\src\CommunityToolkit.WinUI.Controls.Marquee.csproj::TargetFramework=uap10.0.17763]
```
There may have been an issue with the incremental builds when the component was renamed that prevented it from running properly. I'll have to branch off `main` and fix this before we continue.

components/ColorAnalyzer/samples/Assets/StockImages/Bloom.jpg
michael-hawker, line 1: We should use non-copyrighted images here as examples (or at least clearly licensed). Not sure about the others, but grab things from Unsplash and add a note in the docs about the credit to be nice.

components/ColorAnalyzer/src/MultiTarget.props
```diff
@@ -0,0 +1,9 @@
+<Project>
+    <PropertyGroup>
+        <!--
+          MultiTarget is a custom property that indicates which target a project is designed to be built for / run on.
+          Used to create project references, generate solution files, enable/disable TargetFrameworks, and build nuget packages.
+        -->
+        <MultiTarget>uwp;wasdk;</MultiTarget>
```
michael-hawker, line 8: Doesn't work on Uno? Can we add a note here about why?

components/ColorAnalyzer/src/AccentAnalyzer.Properties.cs
```diff
@@ -0,0 +1,152 @@
...
+
+    /// <summary>
+    /// Gets the set of <see cref="AccentColorInfo"/> extracted on last update.
+    /// </summary>
+    public IReadOnlyList<AccentColorInfo>? AccentColors { get; private set; }
+
+    /// <summary>
+    /// Gets a command that executes an accent update.
+    /// </summary>
+    public ICommand AccentUpdateCommand { get; }
```
michael-hawker: Exposing a Command from a control isn't a common pattern. This seems weird. If you want to keep, I'd replace Command implementation from main class with the 'new' `XamlUICommand` base class for this... but otherwise I think should just be a method. App developer can wrap in a command or a behavior to update, if they need to. Is this going to be a common scenario?

components/ColorAnalyzer/src/AccentAnalyzer.Properties.cs
```diff
@@ -0,0 +1,152 @@
...
+
+    /// <summary>
+    /// Gets or sets the <see cref="UIElement"/>
+    /// </summary>
+    public UIElement? Source
```
michael-hawker: Why is this not a DP?

components/ColorAnalyzer/src/AccentAnalyzer.Properties.cs
```diff
@@ -0,0 +1,152 @@
...
+        get => (Color)GetValue(DominantColorProperty);
+        protected set => SetValue(DominantColorProperty, value);
+    }
+    
+    /// <summary>
+    /// Gets the "colorfulness" of the <see cref="Source"/>.
+    /// </summary>
+    /// <remarks>
+    /// Colorfulness is defined by David Hasler and Sabine Susstrunk's paper on measuring colorfulness
+    /// <seealso href="https://infoscience.epfl.ch/server/api/core/bitstreams/77f5adab-e825-4995-92db-c9ff4cd8bf5a/content"/>
```
michael-hawker: Can we provide some examples here of potential small and high values to represent what expected values a developer could expect for certain images? Like is this (black) 0.0-1.0 (white), or is it just an arbitrary number like 44?

michael-hawker (changes requested):
Need at least the images changed I think for licensing concerns.

Future discussion: I could see the calculations done in a .NET library without UI dependency, then maybe a behavior pattern could be better. As basically you could have a 'result' and just regular C# properties without so many DP that way, or you could encapsulate them in another POCO record/class and the result holder is the DP that gets updated and notify bindings and such (I think that'd work...) Anyway, for now this is a good thing to get more feedback on, I know folks have been looking for this type of stuff for a while.

Thanks for the contribution!

components/ColorAnalyzer/src/AccentAnalyzer.Properties.cs
Arlodotexe: This doesn't seem unusual to me, it's the same thing `Button` does.

components/ColorAnalyzer/samples/Assets/StockImages/Bloom.jpg
Avid29: Except for the Windows 11 Bloom which is of course Microsoft, these images are from [Pexels (license info link)](https://www.pexels.com/license/)

components/ColorAnalyzer/src/MultiTarget.props
Avid29: It requires the RenderTargetBitmap, which is not supported on Uno.

Arlo and I have discussed Uno compatible alternatives in instances where we can bind to an image source or its parent, though this is not implemented.

I believe there's a comment by the offending code, but we can add it here too if you'd like.

components/ColorAnalyzer/src/AccentAnalyzer.Properties.cs
Avid29: Good question. I can't think of any reason not to make it one.

components/ColorAnalyzer/src/AccentAnalyzer.Properties.cs
Avid29: We definitely should, but since this property is for the colorfulness of the image and not an individual color that would be challenging.

For the individual colors, it's much easier. I need to do some math to get the value for a color with saturation, but white and black would both be 0

components/ColorAnalyzer/src/MultiTarget.props
Avid29: I am mistaken. I'll add a comment where we use the non-Uno feature and also in this spot

components/ColorAnalyzer/samples/Assets/StockImages/Bloom.jpg
michael-hawker: Cool, I know attribution is not technically required, but it'd still be good for us to support artists and also just know the sources of for future reference to make it easier for us to point to the license in the future.

Can we just swap out the bloom then? 

(@niels9001 I missed that we had added one to the effects gallery in the main repo, we should swap that out too, eh?)

components/ColorAnalyzer/src/AccentAnalyzer.Properties.cs
michael-hawker: Button accepts a command to run when you click it...

https://github.com/CommunityToolkit/Labs-Windows/pull/734/files#diff-cb38a8fa7c92f665bf1b613f8ecf61a22e5d5822d61049c06bbb30db8b0c50e8R50

This is creating a command (in the constructor, link above) intended for developers to run (instead of calling the `UpdateAccent` method), completely opposite... 😅

Plus, with x:Bind now, you should theoretically just be able to call any method on the class instead directly without a command...

components/ColorAnalyzer/src/AccentAnalyzer.Properties.cs
michael-hawker: Yeah, I'm just looking for some general information to help developers understand what this value maps to (like knowing black/white are both zero is a great boundary to call out!). And so then just the opposite side of knowing what a possible 'maximal' or 'normal' value could be. That'll make this more actionable/usable for a developer to understand how/when to use it or what information it actually provides about an image.

components/ColorAnalyzer/src/AccentAnalyzer.Properties.cs
Arlodotexe: Ah, good catch! The pattern with `Button` *is* inverted. We should clean it up and see if x:Bind can be used.

Avid29: The command has been removed. Wasn't sure how to call the function with x:Bind in a behavior, so it does just use `CallMethodAction`

components/ColorAnalyzer/src/AccentAnalyzer.Properties.cs
Avid29: This has been added to the remarks for both `AccentAnalyzer.Colorfulness` and `AccentColorInfo.Colorfulness`

components/ColorAnalyzer/samples/Assets/StockImages/Bloom.jpg
Avid29: Done

michael-hawker (approved):

PR #735, opened 2025-09-25
Marquee State Transition Updates 

Avid29:
Improved API for starting, pausing, resuming, and stopping the Marquee animations

### DIFF
--- a/components/Marquee/samples/Marquee.md
+++ b/components/Marquee/samples/Marquee.md
@@ -50,3 +50,9 @@ The default direction is left, meaning the content will move leftwards, but this
 It is possible to use non-text content in the Marquee control. However templating must be used because the control will need to be duplicated for the looping animation.
 
 > [!Sample MarqueeSample]
+
+# Level Up with Behaviors
+
+Use behaviors to triggers the Marquee on events
+
+> [!Sample MarqueeBehaviorSample]
--- a/components/Marquee/samples/MarqueeBehaviorSample.xaml
+++ b/components/Marquee/samples/MarqueeBehaviorSample.xaml
@@ -1,0 +1,56 @@
+﻿<!--  Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information.  -->
+<Page x:Class="MarqueeExperiment.Samples.MarqueeBehaviorSample"
+      xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+      xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+      xmlns:behaviors="using:CommunityToolkit.WinUI.Behaviors"
+      xmlns:controls="using:CommunityToolkit.WinUI.Controls"
+      xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+      xmlns:interactivity="using:Microsoft.Xaml.Interactivity"
+      xmlns:local="MarqueeExperiment.Samples"
+      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+      mc:Ignorable="d">
+
+    <StackPanel Padding="16">
+        <TextBlock FontWeight="Bold"
+                   Text="This Marquee will loop once when hovered." />
+        <controls:Marquee x:Name="Loop1Marquee"
+                          Background="Transparent"
+                          Behavior="Ticker"
+                          Content="Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua."
+                          FontSize="18"
+                          RepeatBehavior="1x"
+                          Speed="256">
+            <interactivity:Interaction.Behaviors>
+                <interactivity:EventTriggerBehavior EventName="PointerEntered"
+                                                    SourceObject="{Binding ElementName=Loop1Marquee}">
+                    <interactivity:CallMethodAction MethodName="StartMarquee"
+                                                    TargetObject="{Binding ElementName=Loop1Marquee}" />
+                </interactivity:EventTriggerBehavior>
+            </interactivity:Interaction.Behaviors>
+        </controls:Marquee>
+
+        <TextBlock Margin="0,16,0,0"
+                   FontWeight="Bold"
+                   Text="This Marquee will loop while hovered and pause when not hovered." />
+        <controls:Marquee x:Name="MarqueeControl"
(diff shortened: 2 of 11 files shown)


components/Marquee/samples/MarqueeBehaviorSample.xaml
```diff
@@ -0,0 +1,56 @@
...
+                                                    SourceObject="{Binding ElementName=Loop1Marquee}">
+                    <interactivity:CallMethodAction MethodName="StartMarquee"
+                                                    TargetObject="{Binding ElementName=Loop1Marquee}" />
+                </interactivity:EventTriggerBehavior>
+            </interactivity:Interaction.Behaviors>
+        </controls:Marquee>
+
+        <TextBlock Margin="0,16,0,0"
+                   FontWeight="Bold"
+                   Text="This Marquee will loop while hovered and pause when not hovered." />
```
hawkerm: I'd invert this paradigm, so you see it moving when your mouse isn't over it, as an accessibility thing folks are probably going to want to have the text stop moving if they're trying to put their mouse on it.

hawkerm: Ah, we talked on discord, both scenarios have use-cases, maybe add to the docs where and when to use each pattern?

components/Marquee/samples/Marquee.md
hawkerm, line 1: FYI, I noticed the default property values: speed, behaviors/repeat behaviors, and direction are not specified in the docs here.

hawkerm:
Ah, think the PR package has fallen off the feed? Oh, looks like they won't get pushed if a fork?

hawkerm:
FYI, I did use the OpacityMaskView and it worked great to feather the edges (though had to realize I had to set the H/VContentAlignment on the OMV):

```xml
            <controls:OpacityMaskView.OpacityMask>
                <Grid>
                    <Grid.ColumnDefinitions>
                        <ColumnDefinition Width="8" />
                        <ColumnDefinition Width="*" />
                        <ColumnDefinition Width="8" />
                    </Grid.ColumnDefinitions>
                    <Rectangle>
                        <Rectangle.Fill>
                            <LinearGradientBrush StartPoint="0,0" EndPoint="1,0">
                                <GradientStop Offset="0.0" Color="Transparent" />
                                <GradientStop Offset="1.0" Color="White" />
                            </LinearGradientBrush>
                        </Rectangle.Fill>
                    </Rectangle>
                    <Rectangle Grid.Column="1" Fill="White" />
                    <Rectangle Grid.Column="2">
                        <Rectangle.Fill>
                            <LinearGradientBrush StartPoint="0,0" EndPoint="1,0">
                                <GradientStop Offset="0.0" Color="White" />
                                <GradientStop Offset="1.0" Color="Transparent" />
                            </LinearGradientBrush>
                        </Rectangle.Fill>
                    </Rectangle>
                </Grid>
            </controls:OpacityMaskView.OpacityMask>
```

So, this works with the current version, though I'd still expect it to clip to the bounds on its own otherwise you see the text covering other elements and just disappearing all at once.

components/Marquee/samples/Marquee.md
Avid29: Fixed 👍

components/Marquee/samples/MarqueeBehaviorSample.xaml
Avid29: I actually think the `StartAnimation` on hover seen above makes more sense for the use case I was suggesting. I've replaced this sample with an inverted sample containing Hyperlinks.

components/Marquee/src/Marquee.cs
```diff
@@ -126,48 +179,125 @@ public void StartMarquee()
...
     /// </summary>
-    /// <param name="resume">True if animation should resume from its current position, false if it should restart.</param>
+    /// <remarks>
+    /// When in looping mode, it is possible that no animation is necessary.
+    /// </remarks>
+    /// <param name="seekPoint">The seek point to resume the animation (if possible or appropriate.</param>
     /// <exception cref="InvalidOperationException">Thrown when template parts are not supplied.</exception>
-    /// <returns>True if the Animation is now playing.</returns>
-    private bool UpdateAnimation(bool resume = true)
+    /// <returns>Returns whether or not an animation is neccesary.</returns>
```
Arlodotexe:
```suggestion
    /// <returns>Returns whether or not an animation is necessary.</returns>
```

components/Marquee/samples/MarqueeTextSample.xaml
```diff
@@ -8,12 +8,24 @@
...
+        <controls:Marquee x:Name="MarqueeControl"
+                          AutoPlay="{x:Bind MQAuto, Mode=OneWay}"
+                          Behavior="{x:Bind ConvertStringToMarqueeBehavior(MQBehavior), Mode=OneWay}"
                           Content="{x:Bind MQText, Mode=OneWay}"
                           Direction="{x:Bind ConvertStringToMarqueeDirection(MQDirection), Mode=OneWay}"
                           FontSize="18"
                           RepeatBehavior="Forever"
                           Speed="{x:Bind MQSpeed, Mode=OneWay}" />
+
+        <Button Click="StartMarquee_Click"
```
Arlodotexe, line 22: We should move these to an options pane in a future PR. Code-behind event handlers could be kept in-sample since it demonstrates the Start/Stop/Pause/Resume methods, but this XAML is noise.

components/Marquee/src/Marquee.cs
```diff
@@ -204,18 +334,18 @@ _segment1 is null ||
...
-            // calling update animation. If _isActive were passed, it would allow for
-            // MarqueeStopped to be invoked when the marquee was already stopped.
-            StopMarquee(resume);
+            // Reset the transform to 0 and hide the second segment
+            _marqueeContainer.SetValue(dp, 0);
             _segment2.Visibility = Visibility.Collapsed;
-            
+
+            _marqueeStoryboard = null;
+            _marqueeStoryboard?.Stop();
```
Arlodotexe, line 341: Line 341 and 342 need to be swapped, the first makes the second do nothing.

components/Marquee/src/Marquee.cs
```diff
@@ -228,6 +358,8 @@ _segment1 is null ||
         // If the distance is zero, don't play an animation
         if (distance is 0)
         {
+            _marqueeStoryboard = null;
+            _marqueeStoryboard?.Stop();
```
Arlodotexe, line 361: Line 361 and 362 need to be swapped, the first makes the second do nothing.

PR #744, opened 2025-10-02
Use release events to trigger builds from automated weekly releases

Arlodotexe:
## Problem

Automated weekly releases created by `scheduled-releases.yml` were not triggering the build workflow, resulting in missing release artifacts and malformed version numbers.

**Root Cause:** GitHub Actions tokens cannot trigger workflows on tag creation events. When the scheduled workflow creates tags via `git tag` and `git push`, the resulting tag event is ignored by other workflows due to GitHub's security model.

Reference: https://github.com/orgs/community/discussions/76402

> "If an action uses a token to do something to the repo, it won't trigger other actions"

## Solution

Replace `git tag` + `git push` with `gh release create`, which triggers **release events** that CAN be activated by Actions tokens.

**Key Insight:** `gh release create` automatically creates the associated tag if it doesn't exist, so we don't need separate tag creation commands.

### `.github/workflows/scheduled-releases.yml`
- **Removed:** Git configuration (no longer needed with `gh` CLI)
- **Changed:** Tag creation approach from `git tag -a` + `git push` to `gh release create`
- **Added:** Release title and auto-generated release notes
- **Added:** `--prerelease` flag to mark weekly releases appropriately

**Before:**
```yaml
- name: Create and push tag
  run: |
    TAG="release/weekly/$(date +%y%m%d)"
    git config user.name "github-actions[bot]"
    git config user.email "github-actions[bot]@users.noreply.github.com"
    git tag -a "$TAG" -m "Release $TAG"
    git push origin "$TAG"
```

**After:**
```yaml
- name: Create weekly release
  env:
    GH_TOKEN: ${{ github.token }}
  run: |
    TAG="release/weekly/$(date +%y%m%d)"
    gh release create "$TAG" \
      --title "Weekly Release $(date +%y%m%d)" \
      --generate-notes \
      --prerelease
```

### `.github/workflows/build.yml`
- **Added:** `release: types: [published]` trigger to catch automated releases
- **Added:** Filter step to restrict builds to weekly releases only (blocks manual releases with non-matching patterns)

**New trigger:**
```yaml
on:
  push:
    branches: [ main ]
    tags: [ 'release/weekly/**' ]  # Keep for backward compatibility
  pull_request:
    types: [opened, reopened, synchronize]
  release:                          # NEW
    types: [published]
```

**New filter step:**
```yaml
- name: Check if weekly release
  if: github.event_name == 'release'
  shell: pwsh
  run: |
    $tagName = "${{ github.event.release.tag_name }}"
[...]

PR #754, opened 2025-10-16
Fix scheduled release CI trigger and add sparse release notes

Arlodotexe:
Attempts to fix the issue where scheduled weekly releases don't trigger CI builds.

## Problem

Release events created by workflows using `GITHUB_TOKEN` don't trigger other workflows (GitHub Actions security policy to prevent recursive workflow triggering).

## Solution

1. **Remove explicit token references** - Let GitHub Actions handle authentication automatically instead of explicitly setting `GH_TOKEN` or `GITHUB_TOKEN` in env
2. **Add `--notes-start-tag`** - Generate sparse release notes comparing only against previous weekly tag (prevents pinging everyone with full history)
3. **Add `--fail-on-no-commits`** - Prevent empty releases when no work has been done since last release

## Changes

- Modified `.github/workflows/scheduled-releases.yml`
- Removed `env: GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}` block
- Added logic to find previous weekly tag
- Added `--notes-start-tag "$PREV_TAG"` to `gh release create` command
- Added `--fail-on-no-commits` flag

## Testing

Can be tested via:
- Manual trigger: `workflow_dispatch` on Actions tab
- Automated: Wait for next Wednesday's scheduled run (Oct 23)

## Related

- Issue: Scheduled releases not triggering CI builds (discovered Oct 9 sync)
- PR #744 - Original release trigger fix attempt (switched from tag to release events, but didn't trigger release build)

See also
- https://cli.github.com/manual/gh_release_create
- https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=published#release
- https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#triggering-a-workflow-from-a-workflow
- https://docs.github.com/en/actions/tutorials/authenticate-with-github_token

### DIFF
--- a/.github/workflows/scheduled-releases.yml
+++ b/.github/workflows/scheduled-releases.yml
@@ -18,21 +18,23 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
-          # Use a token with write permissions to push to the branch
-          token: ${{ secrets.GITHUB_TOKEN }}
-          fetch-depth: 0 # Fetch all history for merging
+          fetch-depth: 0 # Fetch all history for tags
 
       - name: Create release
         id: create-release
         run: |
           # Date in the format YYMMDD
           TAG="release/weekly/$(date +%y%m%d)"
+          
+          # Get previous weekly tag for sparse release notes
+          PREV_TAG=$(git tag --list 'release/weekly/*' --sort=-creatordate | head -n 1)
+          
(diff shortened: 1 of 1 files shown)


michael-hawker (commented):
See if this works... 🤞

PR #757, opened 2025-10-28
Fix CI manual tag versioning to use tag creation date

Arlodotexe:
## Summary

Fixes issue where manually created tags produce packages with incorrect dates. Tag `release/weekly/251027` (created 10/27) was producing packages dated `251021` (6-day mismatch).

## Root Cause

The CI workflow used `git log -1 --format=%cd` without a tag reference, which extracts the **commit date** instead of the date from the tag name for tagged builds.

### Tag Builds (Fixed)
- Now extracts date directly from tag name using string replacement: `refs/tags/release/weekly/YYMMDD` → `YYMMDD`
- Only sets `VERSION_DATE` environment variable for tag builds
- Allows intentional back-dating or post-dating of releases by using the date in the tag name
- Example: Tag `refs/tags/release/weekly/251027` → packages dated `251027`

### Non-Tag Builds (Main/PR - No Change)
- `VERSION_DATE` environment variable is not set
- Build script defaults to current date via `(Get-Date -UFormat %y%m%d)`
- Behavior unchanged: packages are dated by build time

### Build Script Invocation
- `-DateForVersion` parameter now conditionally passed only when `VERSION_DATE` is set
- Uses GitHub Actions expression: `${{ env.VERSION_DATE != '' && format('-DateForVersion {0}', env.VERSION_DATE) || '' }}`

## Implementation

```yaml
- name: Format Date/Time for Package Version (Tag builds only)
  if: ${{ startsWith(github.ref, 'refs/tags/') }}
  run: |
    # Extract date from tag name (e.g., refs/tags/release/weekly/251027 -> 251027)
    # Non-tag builds: VERSION_DATE not set, script defaults to current date
    $dateString = "${{ github.ref }}" -replace '^refs/tags/release/weekly/', ''
    echo "VERSION_DATE=$dateString" >> $env:GITHUB_ENV
```

## Rationale

Extracting from the tag name (rather than tag creation timestamp) allows for intentional back-dating or post-dating of releases, matching the pattern used in the main CommunityToolkit/Windows repository for `rel/*` branches.


.github/workflows/build.yml
```diff
@@ -287,9 +287,12 @@ jobs:
...
           fetch-tags: true
           
-      - name: Format Date/Time of Commit for Package Version
+      - name: Format Date/Time for Package Version (Tag builds only)
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
         run: |
-          echo "VERSION_DATE=$(git log -1 --format=%cd --date=format:%y%m%d)" >> $env:GITHUB_ENV
+          # Tag builds: use tag creation date (allows manual tags with specific dates)
+          # Non-tag builds: VERSION_DATE not set, script defaults to current date if needed
+          echo "VERSION_DATE=$(git for-each-ref --format='%(creatordate:format:%y%m%d)' ${{ github.ref }})" >> $env:GITHUB_ENV
```
michael-hawker: I'd prefer we just grab the date from the string of the `refs/tags/*` itself, as we don't know if the tag will be created at the same time, we may want to back or post date a release. That's what we do with the `rel/*` branch pattern in the main repo:

https://github.com/CommunityToolkit/Windows/blob/fae89405e770e8a697acabf9045b9b8512df0356/.github/workflows/build.yml#L232-L238

I see us doing the same thing here, just with the tag.

michael-hawker (approved):
LTGM

PR #761, opened 2025-12-01
MarkdownTextBlock improvements

niels9001:
- Bugfixes: certain properties, like the `MinWidth`, `MinHeight` and `Stretch` for `Image` were not working because the code-behind was not leveraging the actual theme, but only listening to the default theme. As a result, setting these properties did nothing.
- Bugfix: `ListBulletSpacing` wasn't working.
- Added `BoldFontWeight` as a property; this overrides the `FontWeight` used for bolded text, default is `Bold` but in the Fluent Design typeramp we usually use `SemiBold` instead. With this property apps can override this to make sure their markdown text renders nicely inline with the rest of the UI typeramps.
- A new sample was added to show a fully custom config that can be customized on the fly:

⚠️ **99% of this code was AI generated by `Claude Opus 4.5` in VS Code.**

image

### DIFF
--- a/components/MarkdownTextBlock/samples/MarkdownTextBlockCustomThemeSample.xaml.cs
+++ b/components/MarkdownTextBlock/samples/MarkdownTextBlockCustomThemeSample.xaml.cs
@@ -84,9 +84,27 @@ The line above is a horizontal rule with customizable thickness and margin.
 
 ## Lists
 
+Try adjusting the **Bullet Spacing** and **Gutter Width** settings to see how list formatting changes!
+
+- Top level list item
+  - Nested item level 1
+    - Nested item level 2
+      - Nested item level 3
+- Another top level item
 - Adjust the theme settings in the options panel
+  - The gutter width controls how much each level is indented
+  - The bullet spacing controls space after the bullet character
 - Click **Apply Changes** to see updates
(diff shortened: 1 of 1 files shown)


Arlodotexe:
> ⚠️ 99% of this code was AI generated by Claude Opus 4.5 in VS Code.

It reinvented the entire sample options pane from scratch and polluted the main sample source code which is displayed to users in the gallery. Instead, it should have:
1. Created a separate control just for the options pane
2. Placed a custom sample options pane attribute on it and
3. Used the `XamlNamedPropertyRelay` to access `x:Name`'d XAML elements in the sample from the sample options pane code-behind. 

Existing samples using this setup can be found in the [ColorAnalyzer](https://github.com/CommunityToolkit/Labs-Windows/blob/main/components/ColorAnalyzer/samples/ContrastHelper/ContrastOptionsPane.xaml.cs) and [Media](https://github.com/CommunityToolkit/Windows/blob/main/components/Media/samples/EffectAnimationsSampleOptions.xaml.cs) components. This is necessary to keep it easy for devs to copy sample source code from the sample gallery app.

niels9001:
@Arlodotexe See latest changes, it now uses a separate UC following what ColorAnalyzer did.

PR #764, opened 2025-12-03
WrapPanel2 Component Creation 

### LINKED ISSUE #763: 🧪 [Experiment] WrapPanel2
### Approved from Discussion

https://github.com/CommunityToolkit/Labs-Windows/discussions/762

### Problem Statement

This panel is an opportunist proposal to expand the `EqualPanel` during discussions of moving it into `Controls.Primitives` and out of `Segmented`.

The panel can combine the roles of the `EqualPanel` **and** the `WrapPanel` into one versatile panel that can both dynamically size children, as well as wrap into new rows/columns.

Considering the default behavior matches the `WrapPanel`, the current name is `WrapPanel2`.

### Overview

This experiment adds a new `WrapPanel2` component to the lab.

Ultimately, this panel is intended to make its way into `Primitives` for the mainline toolkit, as a replacement for the `WrapPanel` and `EqualPanel`.

### Using

You can try it out via the NuGet Packages here:
 - UWP: https://dev.azure.com/dotnet/CommunityToolkit/_artifacts/feed/CommunityToolkit-Labs/NuGet/CommunityToolkit.Labs.Uwp.WrapPanel2/overview/0.0.1
 - WinUI 3: https://dev.azure.com/dotnet/CommunityToolkit/_artifacts/feed/CommunityToolkit-Labs/NuGet/CommunityToolkit.Labs.WinUI.WrapPanel2/overview/0.0.1
Read more about [Preview Packages here](https://aka.ms/wct/wiki/previewpackages).

### Additional info

_No response_

### Implementation Requirements

- [x] Working Prototype
- [x] Feature Complete
- [x] Documentation
- [x] Samples
- [ ] Unit Tests
- [ ] Community Feedback / Usage Testimonies

### Tested Platforms

- [ ] UWP
- [x] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Technical Review

- [ ] Accessibility Audit
- [ ] API/Naming Review
- [ ] Code Quality/Style
- [ ] Dependency Review
- [ ] Design/Style Review
- [ ] Final Approval

### Community Help?

None

Avid29:
Introduces the first draft of the `WrapPanel2`.

Resolves: #763 as discussed in #762.


components/WrapPanel2/samples/WrapPanel2.md
```diff
@@ -0,0 +1,54 @@
...
+
+> [!Sample WrapPanel2BasicSample]
+
+## Properties
+
+### Fixed Row Length
+
+When `FixedRowLengths` is enabled, all rows/columns will to stretch to the size of the largest row/column in the panel. When this is not enabled, rows/columns will size to their content individually.
+
+### Forced Stretch Method
```
michael-hawker: This is old name still?

components/WrapPanel2/samples/WrapPanel2.md
```diff
@@ -0,0 +1,54 @@
...
+# WrapPanel2
+
+The WrapPanel2 is an experiment for a new WrapPanel API using GridLength definitions to define the item's desired sizings.
+
+When stretched along the main axis, the child elements with star-sized GridLength values will proportionally occupy the available space.
+
+When not stretched along the main axis, star-sized child elements will be the smallest size possible while maintaining proportional sizing relative to each other and ensuring that all child elements are fully visible.
+
+
+> [!Sample WrapPanel2BasicSample]
```
michael-hawker, line 27: Thinking it could be good to also split out some more dedicated smaller/fixed examples that show interesting/dedicated uses. The uber example is fun to play with, but I think makes it hard to understand where to use each type of functionality.

i.e. the idea is that the panel is flexible to use in a variety of scenarios, but it'd be rare to try and use every aspect at once.

michael-hawker (commented):
Thanks @Avid29, great start! I just dropping a couple of quick comments.

There's a few option combinations/scenarios that I may be a bit confused on still with how stretching works, maybe it has to do when both children in a row have a specific fixed size, as it's weird to see them not always be that fixed size. Though I know the video is a little out-of-date too, so not sure if that's a factor.

Not sure if it needs to be part of the doc or some other aspect, but I think it'll be good for us to enumerate a bunch of various settings and scenarios and how we expect their behavior to react to layout and see how that matches up with our expectations on various settings and such. Then we can encode those results in the docs/examples/tests.

(This can be an ongoing process too, as we gather feedback.)

components/WrapPanel2/samples/WrapPanel2.md
Avid29: Nice catch. Just updated the markdown

components/WrapPanel2/samples/WrapPanel2.md
Avid29: Yup. I'll change the "Basic Sample" page to "Basic Samples" and make the current sample a "Mega Sample" page.

Avid29:
Fixed a bug with Equal children stretching.

Before:
image

After:
image

components/WrapPanel2/src/WrapPanel2.Properties.cs
```diff
@@ -0,0 +1,142 @@
...
+    public static readonly DependencyProperty ItemSpacingProperty = DependencyProperty.Register(
+        nameof(ItemSpacing),
+        typeof(double),
+        typeof(WrapPanel2),
+        new PropertyMetadata(default(double), OnPropertyChanged));
+
+    /// <summary>
+    /// Backing <see cref="DependencyProperty"/> for the <see cref="LineSpacing"/> property.
+    /// </summary>
+    public static readonly DependencyProperty lineSpacingProperty = DependencyProperty.Register(
```
michael-hawker: missed first capital letter

components/WrapPanel2/src/StretchChildren.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    /// <summary>
+    /// In a row without star-sized items, each item will be stretched to an equal size to fill the row.
+    /// </summary>
+    Equal,
+
+    /// <summary>
+    /// In a row without star-sized items, each item will be stretched proportional to their desired size to fill the row.
+    /// </summary>
+    Proportional,
```
michael-hawker: From discussion here: https://github.com/microsoft/microsoft-ui-xaml/pull/10858#discussion_r2615447331

We're updating to `ItemsStretch` to match UniformGridLayout. Their enum has `Uniform` and `Fill` which I think would be analogs to these defined here, which could be good to match.

Still trying to think about how we reconcile `None` in the base implementation (i.e. so that this was purely a drop-in replacement when stable). My gut is that `None` is the `StarSizedOnly` that we don't have to change that as overriding the attached property to get that behavior is the change in behavior itself... (e.g. in most games, you follow the default rules unless told otherwise, the attached property is a told otherwise extension here basically... [even if the panel is the thing that has to be aware in the implementation, that's an implementation detail])

components/WrapPanel2/src/WrapPanel2.Properties.cs
```diff
@@ -0,0 +1,142 @@
...
+    public double LineSpacing
+    {
+        get => (double)GetValue(lineSpacingProperty);
+        set => SetValue(lineSpacingProperty, value);
+    }
+
+    /// <summary>
+    /// Gets or sets whether or not all rows/columns should stretch to match the length of the longest.
+    /// </summary>
+    public bool FixedRowLengths
```
michael-hawker: Is this the equivalent of [`ItemsJustification`](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.uniformgridlayout.itemsjustification), is that the direction we should move toward?

components/WrapPanel2/src/WrapPanel2.Properties.cs
```diff
@@ -0,0 +1,142 @@
...
+    {
+        get => (bool)GetValue(FixedRowLengthsProperty);
+        set => SetValue(FixedRowLengthsProperty, value);
+    }
+
+    /// <summary>
+    /// Gets or sets the method used to fill rows without a star-sized item.
+    /// </summary>
+    public StretchChildren StretchChildren
```
michael-hawker: See discussion here: https://github.com/microsoft/microsoft-ui-xaml/pull/10858#discussion_r2615447331

We're updating to `ItemsStretch` to match UniformGridLayout.

components/WrapPanel2/src/OverflowBehavior.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// When an item exceeds the available space, it will be moved to a new row or column.
+    /// </summary>
+    Wrap,
+
+    /// <summary>
+    /// Items which do not fit within the available space will be removed from the layout.
+    /// </summary>
+    Drop,
```
michael-hawker: Do we have a scenario/need for this? I'm just curious as it's adding a whole extra property.

michael-hawker (commented):
Some initial comments based on the WrapPanel WinUI API review/direction.

components/WrapPanel2/src/WrapPanel2.Properties.cs
Avid29: My only hesitation was that "justification" is defined by stretching items to touch the margin on both ends. Technically this is not the case unless the main axis stretch is also "Stretch".

Although, since this property applies to the arrange step and not the measure step, arguably the margin is the measured size.

I'll add that rename to my commit addressing your comments, but I'm going to stick with a boolean for now. Especially since that enum seems to have two values with the same description?

components/WrapPanel2/src/OverflowBehavior.cs
Avid29: I had a potential use case when I was writing this, but I wasn't actually happy with it when I plugged it in. I'll remove the property.

components/WrapPanel2/src/StretchChildren.cs
Avid29: Hmm. I see what you mean. I suppose the proper thing to do is name according to the simplest use case and document the extraordinary case as plainly as possible.

Avid29: This is making me think I should also change the behavior of the `ItemJustification` property to adjust the margins when `ItemsStretch` is `None`. I'll do that in another commit after I address the simple changes. I'm actually really happy with that addition to the API as well.

Avid29:
Here's the new behavior for when `ItemJustification` is enabled and `ItemsStretch` is `None`.
image

It will override the `ItemSpacing` to ensure the first and last item are aligned with the margins (if there is only one item it will align to the left margin)

components/WrapPanel2/src/WrapPanel2.Properties.cs
michael-hawker: FYI, enum is here has 6 values: https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.uniformgridlayoutitemsjustification - even if we only have two for now an enum allows us to add more or change it more easily in the future than a bool.

Avid29:
> Cool, `ItemSpacing` then becomes the minimum as a constraint for layout then in the justification scenario, makes sense! (not sure the best place to call that out on which property, maybe ItemSpacing itself)

Hmm. Maybe worth leaving a remark on both properties, but definitely should be a remark for `ItemsSpacing`. I think it's through enough for `ItemsJustification`to just say it forces items to span from margin to margin. It uses a lot of different strategies to do so depending on other properties.

components/WrapPanel2/src/WrapPanel2.Properties.cs
Avid29: Right, I see that. I'm just not sure what the difference between `SpaceEvenly` and `SpaceAround` is. They have the same description.

components/WrapPanel2/samples/WrapPanel2.md
```diff
@@ -0,0 +1,67 @@
...
+icon: assets/icon.png
+---
+
+# WrapPanel2
+
+The WrapPanel2 is an experiment for a new WrapPanel API using GridLength definitions to define the item's desired sizings.
+
+When stretched along the main axis, the child elements with star-sized GridLength values will proportionally occupy the available space.
+
+When not stretched along the main axis, star-sized child elements will be the smallest size possible while maintaining proportional sizing relative to each other and ensuring that all child elements are fully visible.
```
michael-hawker: Could be good to clarify what the default behavior is here and how it compares to the standard WrapPanel.

Also not clear to me what "stretched along the main axis" means, is that setting a specific property to a specific value? If so, we should call that out more explicitly.

components/WrapPanel2/src/WrapPanel2.Properties.cs
```diff
@@ -0,0 +1,137 @@
...
+
+    /// <summary>
+    /// Gets or sets whether or not all rows/columns should stretch its children to ensure the space is filled from margin to margin.
+    /// </summary>
+    /// <remarks>
+    /// This will not apply on lines without star-sized items unless a <see cref="ItemsStretch"/> behavior is selected.
+    /// </remarks>
+    public WrapPanelItemsJustification ItemsJustification
+    {
+        get => (WrapPanelItemsJustification)GetValue(ItemsJustificationProperty);
```
michael-hawker, line 104: Got a crash here trying to load the sample page when running from the component project:

Image

Issue with demo attributes maybe?

components/WrapPanel2/samples/WrapPanel2BasicSample.xaml.cs
```diff
@@ -0,0 +1,118 @@
...
+[ToolkitSampleMultiChoiceOption("LayoutItemsJustification", "Automatic", "Start", "Center", "End", "SpaceAround", "SpaceBetween", "SpaceEvenly", Title = "Items Justification")]
+[ToolkitSampleMultiChoiceOption("LayoutItemsStretch", "None", "First", "Last", "Equal", "Proportional", Title = "Items Stretch")]
+
+[ToolkitSample(id: nameof(WrapPanel2BasicSample), $"Basic demo of the {nameof(WrapPanel2)} with auto-sized items.", description: $"A sample showing every property of the {nameof(WrapPanel2)} panel.")]
+public sealed partial class WrapPanel2BasicSample : Page
+{
+    public WrapPanel2BasicSample()
+    {
+        this.InitializeComponent();
+    }
```
michael-hawker, line 26: We should start with some items in the collection by default, I was very confused about the properties not doing anything at first, not realizing the buttons were control buttons and not part of the collection.

We actually need to just make these action buttons consistent with a little docked status bar or something in the sample infrastructure, it's been on the backlog for a while, maybe I'll take a quick look into: https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/4

components/WrapPanel2/src/WrapPanel2.Properties.cs
```diff
@@ -0,0 +1,137 @@
...
+        set => SetValue(ItemsJustificationProperty, value);
+    }
+
+    /// <summary>
+    /// Gets or sets the method used to fill rows without a star-sized item when <see cref="ItemsJustification"/> is in a spacing mode.
+    /// </summary>
+    public WrapPanelItemsStretch ItemsStretch
```
michael-hawker, line 111: Hmm, I wasn't seeing these do much unless the justification was set to one of the spaced properties. May need to think about the behaviors of some of the combinations here. I need to play with this a bit more.

michael-hawker (commented):
Some quick initial comments, but need to play a bit more with it, just had a few minutes to start digging into it.

components/WrapPanel2/src/WrapPanel2.Properties.cs
```diff
@@ -0,0 +1,137 @@
...
+
+namespace CommunityToolkit.WinUI.Controls;
+
+public partial class WrapPanel2
+{
+    /// <summary>
+    /// An attached property for identifying the requested layout of a child within the panel.
+    /// </summary>
+    public static readonly DependencyProperty LayoutLengthProperty =
+        DependencyProperty.Register(
```
michael-hawker:
```suggestion
        DependencyProperty.RegisterAttached(
```
Attached Property should use RegisterAttached method

components/WrapPanel2/src/WrapPanel2.Properties.cs
```diff
@@ -0,0 +1,137 @@
...
+        new PropertyMetadata(default(double), OnPropertyChanged));
+
+    /// <summary>
+    /// Backing <see cref="DependencyProperty"/> for the <see cref="ItemsJustification"/> property.
+    /// </summary>
+    public static readonly DependencyProperty ItemsJustificationProperty = DependencyProperty.Register(
+        nameof(ItemsJustification),
+        typeof(WrapPanelItemsJustification),
+        typeof(WrapPanel2),
+        new PropertyMetadata(default(bool), OnPropertyChanged));
```
michael-hawker:
```suggestion
        new PropertyMetadata(default(WrapPanelItemsJustification), OnPropertyChanged));
```

Mismatched property type in registration

michael-hawker (commented):
Looks like I can't push to this branch (well maybe, but not how I checked it out), these were the two spots I had to fix for the third demo to work.

components/WrapPanel2/src/WrapPanel2.Properties.cs
michael-hawker: Found issue, applying fix.

michael-hawker:
Been working on https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/4 to clean-up the interface a bit, I was confused at first (we should have a few items in the collection by default for the basic sample). Going to open PR with initial support for that, but we can do that independently of this PR, it's a small update later.

components/WrapPanel2/src/CommunityToolkit.WinUI.Controls.WrapPanel2.csproj
michael-hawker, line 1: Should this be `CommunityToolkit.WinUI.Controls.WrapPanel2.csproj` for now, as it's the only control in this package?

components/WrapPanel2/samples/WrapPanel2BasicSample.xaml.cs
michael-hawker:
```suggestion
        Add5ItemsClick(null!, null!);
    }
```
Should we just call the method for now in the constructor to have something?

components/WrapPanel2/src/WrapPanel2.Properties.cs
michael-hawker: Yeah, played with this again, unless I had `HorizontalAlignment` set to `Stretch` the `ItemsStretch` wasn't doing a lot. I wonder if we need to distinguish these scenarios a big more. Could be good to get some input from a broader set of folks. So, let's ensure we get this merged and move forward with some discussions on expected behaviors/settings combinations (there's a lot now with this!)

components/WrapPanel2/samples/WrapPanel2.md
```diff
@@ -0,0 +1,92 @@
...
+subcategory: Panel
+discussion-id: 762
+issue-id: 763
+icon: assets/icon.png
+---
+
+# WrapPanel2
+
+The WrapPanel2 is an advanced layout control that uses `GridLength` definitions to manage item sizing within a wrapping flow. It provides granular control over how items occupy space, particularly when using proportional (Star) sizing.
+
```
michael-hawker, line 18: Could be good to add more here about how this extends/differs from the original WrapPanel.

(We should decide if one of our goals is drop-in upgrade or not, but that can come later.)

michael-hawker (changes requested):
Couple small things. I think the main thing to ensure we fix is just the project name for the package, should we align it to the control name? (Or would we expect other panels to put in here, then should we make it plural now?)

Once we fix that, we can merge. Looks good! Thanks! 🦙❤️

michael-hawker (approved):
Thanks @Avid29! 🦙❤️ Let's merge this in and fix anything else forward!

PR #767, opened 2025-12-24
Add ColorSource abstraction for the origin of a ColorAnalyzer pixel stream

Avid29:
This PR adds a `ColorSource` type to allow methods of sourcing pixels for the `ColorAnaylzer` besides just rendertarget bitmaps on UIElements


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "9.0.101",
+    "version": "10.0.101",
```
Arlodotexe: @Avid29 Could you add some details on why this was bumped here? It will affect other components and the CI setup if not done carefully.

Arlodotexe (approved):

global.json
Avid29: Oh. When I downloaded VS 2026 it uninstalled the .NET 9 SDK. In order to build on my machine I need to bump this number.

I didn't mean to push this. I must've accidently staged it.

PR #769, opened 2026-01-08
GradientSlider component creation

### LINKED ISSUE #768: 🧪 [Experiment] GradientSlider
### Approved from Discussion

https://discord.com/channels/372137812037730304/1458607373420597260/1458607376914714704

### Problem Statement

Allow easily user modified color gradients.

The inspiration in this case was as a tool for XAML studio, where a user might want to modify a `LinearGradientBrush`, but it could also apply for art or image editing software.

### Overview

This experiment adds a new `GradientSlider` control and a `GradientSliderThumb` for the gradient stops inside the slider. 

### Using

You can try it out via the NuGet Packages here:
 - UWP: https://dev.azure.com/dotnet/CommunityToolkit/_artifacts/feed/CommunityToolkit-Labs/NuGet/CommunityToolkit.Labs.Uwp.ExperimentName/overview/0.0.1
 - WinUI 3: https://dev.azure.com/dotnet/CommunityToolkit/_artifacts/feed/CommunityToolkit-Labs/NuGet/CommunityToolkit.Labs.WinUI.ExperimentName/overview/0.0.1
Read more about [Preview Packages here](https://aka.ms/wct/wiki/previewpackages).

### Additional info

Original inspiration photo (screenshot from Blender)

Image

### Implementation Requirements

- [x] Working Prototype
- [ ] Feature Complete
- [ ] Documentation
- [ ] Samples
- [ ] Unit Tests
- [ ] Community Feedback / Usage Testimonies

### Tested Platforms

- [ ] UWP
- [x] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Technical Review

- [ ] Accessibility Audit
- [ ] API/Naming Review
- [ ] Code Quality/Style
- [ ] Dependency Review
- [ ] Design/Style Review
- [ ] Final Approval

### Community Help?

None

Avid29:
Introduces the first draft of the `GradientSlider` component.

Resolves: #768

https://github.com/user-attachments/assets/d448d35f-5dfb-4fb7-8c01-a874f71178a1


components/GradientSlider/samples/GradientSliderSample.xaml
```diff
@@ -0,0 +1,23 @@
...
+    <StackPanel Spacing="16">
+        <controls:GradientSlider>
+            <controls:GradientSlider.GradientStops>
+                <GradientStopCollection>
+                    <GradientStop Offset="0" Color="#FF0000" />
+                    <GradientStop Offset="0.3" Color="#00FF00" />
+                    <GradientStop Offset="0.66" Color="#0000FF" />
+                    <GradientStop Offset="1" Color="#FF0000" />
+                </GradientStopCollection>
+            </controls:GradientSlider.GradientStops>
```
michael-hawker, line 20: What happens if the developer doesn't specify any stops at the start? Could be good to have another example where it's bound to a collection (to show binding) but that collection is also empty to start?

components/GradientSlider/src/CommunityToolkit.WinUI.Controls.GradientSlider.csproj
```diff
@@ -0,0 +1,14 @@
...
+    <ToolkitComponentName>GradientSlider</ToolkitComponentName>
+    <Description>This package contains GradientSlider.</Description>
+
+    <!-- Rns suffix is required for namespaces shared across projects. See https://github.com/CommunityToolkit/Labs-Windows/issues/152 -->
+    <RootNamespace>CommunityToolkit.WinUI.Controls.GradientSliderRns</RootNamespace>
+  </PropertyGroup>
+
+  <!-- Sets this up as a toolkit component's source project -->
+  <Import Project="$(ToolingDirectory)\ToolkitComponent.SourceProject.props" />
+</Project>
```
michael-hawker:
```suggestion
  <Import Project="$(ToolingDirectory)\ToolkitComponent.SourceProject.props" />
  
  <PropertyGroup>
    <PackageId>$(PackageIdPrefix).$(PackageIdVariant).Controls.$(ToolkitComponentName)</PackageId>
  </PropertyGroup>
</Project>
```

We need the `Controls` added to the package, right @Arlodotexe? (Remind me again why we don't grab this from the csproj name?)

components/GradientSlider/src/GradientSlider.cs
```diff
@@ -0,0 +1,159 @@
...
+        // Untrack the thumb and remove from the canvas
+        _stopThumbs.Remove(stop);
+        _containerCanvas.Children.Remove(thumb);
+    }
+
+    private void RefreshThumbs()
+    {
+        ClearThumbs();
+        foreach (var stop in GradientStops)
+            AddStop(stop);
```
michael-hawker:
```suggestion
        foreach (var stop in GradientStops)
        {
            AddStop(stop);
        }
```

Should encase in closure, @Arlodotexe do we not have this in our editor config? Or is it just not enforced?

components/GradientSlider/src/GradientSlider.Properties.cs
```diff
@@ -0,0 +1,37 @@
...
+        set => SetValue(GradientStopsProperty, value);
+    }
+
+    private static void GradientStopsChangedCallback(DependencyObject d, DependencyPropertyChangedEventArgs e)
+    {
+        if (d is not GradientSlider slider)
+            return;
+
+        if (slider._isDragging)
+            return;
```
michael-hawker: So if a stop is somehow programmatically added while the slide is in use, the slider will never get updated and be out-of-sync? We probably need to call RefreshThumbs or some other sync method when the drag is completed, eh?

components/GradientSlider/src/GradientSliderThumbStyle.xaml
```diff
@@ -0,0 +1,126 @@
...
+                                        </DoubleAnimationUsingKeyFrames>
+                                    </Storyboard>
+                                </VisualState>
+                            </VisualStateGroup>
+                        </VisualStateManager.VisualStateGroups>
+                        <FlyoutBase.AttachedFlyout>
+                            <Flyout Placement="Bottom">
+                                <muxc:ColorPicker x:Name="PART_ColorPicker" />
+                            </Flyout>
+                        </FlyoutBase.AttachedFlyout>
```
michael-hawker, line 108: I'm always leery of adding intertwined children/parents, but I think in this case, it'd be nice if there was a single attachedflyout was part of the parent control and the child invoked that atop itself? Obviously, requires a bit more coordination, but...

Then I think it gets easier to expose the properties of the ColorPicker itself in one place on the GradientSlider. i.e. if they want to show a specific color palette or other things instead, they just have to restyle the parent control which is much simpler than the thumb.

Should we also use the WCT ColorPicker by default instead? Or should we expose a `ColorPicker` property on the slider so the developer themselves provide the picker they want to use configured exactly how they want (we can have a style value set with a default basic one in case they don't provide one)?

components/GradientSlider/src/GradientSlider.cs
michael-hawker, line 1: I would think, double-tapping on the slider should be the mechanism to add a new stop, eh? @niels9001 any suggestions on input here?

I'm trying to think of how you'd remove a stop, would it be dragging it further off some amount the end of the slider?

The only other scenario I'm thinking of here that'd be more complex to add is if you wanted a specific input box for the offset vs. drag input. (It could be nice in either case as you're dragging to see the offset value as a float [0-1] and/or percentage maybe?) Maybe it's not something we provide by default out-of-the-box, but something we should have a sample on how to extend the control to enable. i.e. could the flyout be customized to contain more than just a colorpicker, would we have to search the flyout for the colorpicker instance then?

michael-hawker (commented):
Some initial comments, this is an amazing start! Really interesting use of the Canvas. I'll have to pull it down and play with it a bit more directly.

I'm also wondering since I added Thumbs to the ResizeAdorner, and you have them here if there's any commonalities and if we just need a general `Thumb` base class like I did for the `Sizers` (which could honestly maybe re-use that as well, I mean I basically took that code and made it two-dimensional for the Adorner thumb).

components/GradientSlider/src/GradientSlider.cs
Avid29: It's a suggestion in the editor config. I tend to ignore it when the code on the nested line is shorter than the line above it. Do you have a different rule, since I'm sure you've approved code I've written with this on if cases at least

components/GradientSlider/src/GradientSlider.cs
Avid29: I've added a "Placeholder Thumb" that shows under a hovering mouse. My idea is that clicking there will add a stop in that position.

image

michael-hawker (approved):

components/GradientSlider/src/CommunityToolkit.WinUI.Controls.GradientSlider.csproj
Arlodotexe: Yes, we need the `Controls` part declared here same as we do for other components under this "Controls" package-namespace. 

We don't grab it from the csproj because the `PackageIdVariant` changes depending on whether we're building Uwp or Wasdk, and the csproj name needs to stay the same as the assembly name iirc.

components/GradientSlider/src/GradientSlider.Properties.cs
Arlodotexe: Seconded. This is another known/expected trivial MVP feature, the various properties on this control should support live updates. 

We'll likely need to do a sweep to make sure property change updates are wired everywhere.

components/GradientSlider/samples/GradientSliderSample.xaml
Arlodotexe: Seconded, documenting (and defining if needed) the expected behavior for empty/default properties is a trivial known/expected scenario for any component MVP.

components/GradientSlider/src/GradientSliderThumbStyle.xaml
Arlodotexe: Just to make sure I understand what you're saying, we shouldn't have to retemplate the entire control to swap out the color picker or add any custom content around it? That instead, we could pass it in via a DP deriving ColorPicker, or if we wanted to allow custom content, a Control instance or DataTemplate?

components/GradientSlider/src/GradientSlider.cs
Arlodotexe: I'm a fan of the hover-to-add interaction method, though it doesn't cover removal. We can defer addressing both of these until later, but let's track it in a new ticket after closing this PR.

components/GradientSlider/samples/GradientSlider.md
```diff
@@ -0,0 +1,25 @@
+---
+title: GradientSlider
+author: githubaccount
```
Arlodotexe: Noting missing attribution and metadata in this document

components/GradientSlider/samples/GradientSliderSample.xaml.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using CommunityToolkit.WinUI.Controls;
+
+namespace GradientSliderExperiment.Samples;
+
+/// <summary>
+/// An example sample page of a custom control inheriting from Panel.
+/// </summary>
+
+[ToolkitSample(id: nameof(GradientSliderSample), "Custom control", description: $"A sample for showing how to create and use a {nameof(GradientSlider)} custom control.")]
```
Arlodotexe: Default text (name/description) should be replaced so it's displayed properly in the sample app.

Arlodotexe (approved):
Reviewed and approved, there are improvements to make but nothing we can't defer. Let's take the lingering review comments and turn them into tracking issues, once we've resolved the review threads by deferring them to issues, we can merge the code and close the PR.

components/GradientSlider/src/GradientSlider.Properties.cs
Avid29: It's actually worse than that.

There is no event or callback on the `GradientStopCollection` to detect a stop being added/removed. The actual limitation here is that if the `GradientStops` property values changes (as in the reference to the `GradientStopsCollection` object, while the user is dragging then it will become completely out of sync until the next `RefreshThumbs()` event.

components/GradientSlider/src/GradientSliderThumbStyle.xaml
Avid29: @Arlodotexe Yes, but as a result the `GradientSlider` would own the `ColorPicker`. This has a cost of higher control coupling, but it is more layout optimized and makes this control more easily extensible (even though its components become less extensible).

components/GradientSlider/src/GradientSlider.cs
Avid29: Currently you can right-click to remove a stop. Not the most intuitive solution, but just making clear that there is a way to remove stops right now.

components/GradientSlider/samples/GradientSlider.md
Avid29: Fixed

Avid29:
I've taken care of the simpler problems (default strings and metadata)

michael-hawker:
@Arlodotexe seems like WIP is stuck again... 🤔 I confirmed [our installation](https://github.com/organizations/CommunityToolkit/settings/installations/352986) is this repo: https://github.com/wip/action - they have a status here which shows it's been up: https://stats.uptimerobot.com/Dq46zf6PY - maybe we should file an issue anyway? I'm not sure how we get more details about what's holding it up though (it's a simple action and it hasn't changed).

michael-hawker:
Figured we should just merge this since we've done a lot of the initial review. @Avid29 feel free to open any other tracking issues and such and tie them to the parent experiment. Thanks!

Issue #776, opened 2026-01-20
MarkdownTextBlock not working in multithreaded window


wherewhere:
### Describe the bug

`MarkdownTextBlock` not working in multithreaded window both in UWP and WAS.

```shell
System.Runtime.InteropServices.COMException (0x8001010E)
   at WinRT.ExceptionHelpers.<ThrowExceptionForHR>g__Throw|38_0(Int32 hr)
   at ABI.Windows.UI.Xaml.Documents.ITextElementMethods.set_Foreground(IObjectReference _obj, Brush value)
   at Windows.UI.Xaml.Documents.TextElement.set_Foreground(Brush value)
   at CommunityToolkit.WinUI.Controls.TextElements.MyHeading.SetHProperties(Int32 level)
   at CommunityToolkit.WinUI.Controls.TextElements.MyHeading..ctor(HeadingBlock headingBlock, MarkdownConfig config)
   at CommunityToolkit.WinUI.Controls.Renderers.ObjectRenderers.HeadingRenderer.Write(WinUIRenderer renderer, HeadingBlock obj)
   at Markdig.Renderers.MarkdownObjectRenderer`2.Write(RendererBase renderer, MarkdownObject obj)
   at Markdig.Renderers.RendererBase.Write(MarkdownObject obj)
   at Markdig.Renderers.RendererBase.WriteChildren(ContainerBlock containerBlock)
   at Markdig.Renderers.RendererBase.Write(MarkdownObject obj)
   at CommunityToolkit.WinUI.Controls.Renderers.WinUIRenderer.Render(MarkdownObject markdownObject)
   at CommunityToolkit.WinUI.Controls.MarkdownTextBlock.ApplyText(Boolean rerender)
   at CommunityToolkit.WinUI.Controls.MarkdownTextBlock.Build()
   at CommunityToolkit.WinUI.Controls.MarkdownTextBlock.OnApplyTemplate()
   at Windows.UI.Xaml.FrameworkElement.Windows.UI.Xaml.IFrameworkElementOverrides.OnApplyTemplate()
   at ABI.Windows.UI.Xaml.IFrameworkElementOverrides.Do_Abi_OnApplyTemplate_2(IntPtr thisPtr)
```

### Steps to reproduce

```text
[...]

PR #777, opened 2026-01-20
Fix MarkdownTextBlock which is not working in multithreaded

wherewhere:
See #776

### DIFF
--- a/components/MarkdownTextBlock/src/MarkdownTextBlock.Properties.cs
+++ b/components/MarkdownTextBlock/src/MarkdownTextBlock.Properties.cs
@@ -15,7 +15,7 @@ public partial class MarkdownTextBlock
         nameof(Config),
         typeof(MarkdownConfig),
         typeof(MarkdownTextBlock),
-        new PropertyMetadata(new MarkdownConfig(), OnConfigChanged)
+        new PropertyMetadata(MarkdownConfig.Default, OnConfigChanged)
     );
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


components/MarkdownTextBlock/src/MarkdownTextBlock.Properties.cs
```diff
@@ -15,7 +15,7 @@ public partial class MarkdownTextBlock
         nameof(Config),
         typeof(MarkdownConfig),
         typeof(MarkdownTextBlock),
-        new PropertyMetadata(new MarkdownConfig(), OnConfigChanged)
+        new PropertyMetadata(MarkdownConfig.Default, OnConfigChanged)
```
michael-hawker: The real problem here is we need to be using the CreateDefaultValueCallback for the default value of a reference type, see Docs for DP: https://learn.microsoft.com/windows/apps/develop/platform/xaml/custom-dependency-properties#createdefaultvaluecallback

wherewhere: So it need to be `() => MarkdownConfig.Default` here? But it works in my place...

wherewhere: Replaced to `CreateDefaultValueCallback`.

chawolbaka:
Screenshot_2026-06-29-20-02-54-84_9d26c6446fd7bb8e41d99b6262b17def

PR #779, opened 2026-02-02
Update .NET SDK to 9.0.310 and use global.json for CI version control

Arlodotexe:
## Problem

The current CI setup has version specified in two places:
1. `global.json` with exact version (9.0.308) and `rollForward: disable`
2. `DOTNET_VERSION` env var in workflow with floating version (9.0.x)

This led to a previous breakage when GitHub Actions runners installed 9.0.309 while global.json expected 9.0.308 (fixed in #826 by disabling rollForward).

## Solution

This PR implements single source of truth for .NET SDK version:
- **Version update**: 9.0.308 → 9.0.310
- **Workflow pattern**: Use `global-json-file: global.json` input for `actions/setup-dotnet@v4`
- **Step reordering**: Checkout repository before setup-dotnet (required for global.json access)

## Changes

**global.json**:
- Updated SDK version to 9.0.310

**build.yml**:
- Reordered steps: `actions/checkout` now precedes `actions/setup-dotnet`
- Changed setup-dotnet input from `dotnet-version: 9.0.x` to `global-json-file: global.json`
- Retained `DOTNET_VERSION` env var for artifact-only jobs (sign, nuget) that cannot access global.json

## Benefits

- ✅ Single source of truth for SDK version
- ✅ Prevents future version drift
- ✅ Simplifies version updates (edit global.json only)
- ✅ Follows setup-dotnet recommended pattern

## Related

- #826 - Disabled rollForward to prevent SDK mismatch

### DIFF
--- a/components/MarkdownTextBlock/samples/MarkdownTextBlock.md
+++ b/components/MarkdownTextBlock/samples/MarkdownTextBlock.md
@@ -39,4 +39,3 @@ Try typing markdown and see it rendered in real-time:
 Try different styling options for all markdown elements with a custom theme:
 
 > [!Sample MarkdownTextBlockCustomThemeSample]
-
--- a/components/MarkdownTextBlock/samples/MarkdownTextBlockCustomThemeSample.xaml.cs
+++ b/components/MarkdownTextBlock/samples/MarkdownTextBlockCustomThemeSample.xaml.cs
@@ -40,6 +40,8 @@ Images can be styled with max width, max height, and stretch options. Notice how
 
 The image above automatically scales to respect the max width setting while maintaining its aspect ratio. Text continues to flow seamlessly below it.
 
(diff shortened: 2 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -24,7 +24,8 @@ on:
   merge_group:
 
 env:
-  DOTNET_VERSION: ${{ '9.0.x' }}
+  DOTNET_VERSION: '9.0.310'  # Only used by jobs that cannot access global.json (no checkout/artifact-only jobs)
```
michael-hawker, line 27: In these cases it's only for nuget push, eh? Is there an alternate download/install way for the nuget tool vs. dotnet we could use to avoid tracking this version here?

Arlodotexe: Our options are:
1. What we did here
2. Hardcoding inline
3. Cloning the repo so we have access to global.json

Arlodotexe: Noting also duplicate with https://github.com/CommunityToolkit/Windows/pull/827/#discussion_r2756483423

PR #781, opened 2026-02-19
Improve DataTable control

9rnsr:
- Update layout logic to improve the support quality of mixing Auto, Star, and Absolute sized columns. And, for Auto and Star columns, distinguish the two states between auto-calculated widths and manually resized ones (the later would work like absolute sized columns in layout logic).
- Set `DataColumn.IsTabStop` `false` (an issue in [here](https://github.com/CommunityToolkit/Labs-Windows/discussions/415#discussioncomment-7006205))
- Consider `DataColumn.Visibility` in the layout logic.

And I removed the hybrid-case support (`DataTableHybridSample`). As I commented [here](https://github.com/CommunityToolkit/Labs-Windows/discussions/415#discussioncomment-15855673), it would not work on `Auto` columns as expected, so we should use the pair of `DataTable` and `DataRow` always.


9rnsr:
@michael-hawker I opened independent smaller PRs #782, #783, and #784.

michael-hawker:
Thanks @9rnsr! I have a few other items I'm looking at currently, but the small break-outs will make it easier to start reviewing/discussing these bits independently and start merging them in over the next week!

Just to confirm, we can close this PR now as it'll be superseded by the others?

9rnsr:
> Just to confirm, we can close this PR now as it'll be superseded by the others?

@michael-hawker No problem.

9rnsr:
I maintain and improve the core logic update.

Also I rebase the commits: five preparation changes from the PRs (still opened), two core update commits.

michael-hawker:
Thanks @9rnsr for breaking this up, it certainly helped look at smaller improvements in the other PRs.

I did come back to this one to try and better understand the changes. Behavior wise, it seems to fix a few things, though there's a regression with the `DataTableVirtualizationSample`. If you resize the column first and then scroll, it makes the column larger to fit the content now when it should respect the new fixed size.

I think that's one of my concerns here with this large a change is that there seems to be a lot of extra work done to re-measure/arrange in the layout passes. This fixes some issues like the column jitter and keyboard resizing, but is also pushing a lot of extra layout cycles. I wonder if there's a better happy medium (though there is certainly some of that always going to occur with a column resize from the drag event, but outside that is the main concern)?

I wonder if we can have some tests that either help validate the minimum number of passes required, but also behavior for some of these scenarios could help convey the changes. Also, maybe a few more comments or at least maybe some more annotations here in the PR around why changes were made and what scenarios they're handling.

9rnsr:
> I did come back to this one to try and better understand the changes. Behavior wise, it seems to fix a few things, though there's a regression with the `DataTableVirtualizationSample`. If you resize the column first and then scroll, it makes the column larger to fit the content now when it should respect the new fixed size.

Oops, it's my fault. When I updated the logic once after starting this PR, I was accidentally added the bug. Now it's fixed.

9rnsr:
The core part of my improvement for this `DataTable` control is: **rely on the layout system**.

Let's try to explain the the layout steps for the calculation width of an "auto column" (means `IsAuto == true && IsFixed == false`):

1. First, `DataTable.MeasureOverride` is invoked. If there's any auto columns, all `DataRow` instances will get the measure phase invalidations (see the code pass `invokeRowsMeasures = true`).

2. Next, `DataRow.MeasureOverride` checks and updates `column.ActualColumnWidth` for the auto column. `DataTable` will get the 2nd measure phase invalidation *only when* the column content is wider than the column width.

3. The 2nd `DataTable.MeasureOverride` invokation does same things with `#1` and all `DataRow` instances will get the measure phase invalidations again, because there is no identifiable difference from `#1`.

4. `DataRow.MeasureOverride` checks `column.ActualColumnWidth`, but no one update it so `DataTable` doesn't get more measure phase invalidation. Finally stops layout cycle.

Indeed the *invalidation* calls are increased. However I expect the layout system will elide the actual repeated calculation on the elements, with the cache of previous measurement result to `DesiredSize`, ignore elements out of the viewport, and the stop layout cycle when the measurements and arrangements are reached to the fixed point.

Actually, we can see it with `DataTableVirtualizationSample`, the spreading of "Description" column width is delayed until we move the scroll position to the line 1500. When the data-table displayed (scroll position is top of the view), the line 1500 gets measure phase invalidation in `#1`, but layout system elide `DataRow.MeasureOverride` call because its out of the viewport and `#2`~`#4` doesn't happen.

From an another perspective, the steps `#2` and` #4` looks like a repetition, but all measurement results (the value of `child.Measure()` calls) of the `DataRow` children will not change between the two steps, so the result in `#4` step would be the cached value. In other words, the `#2` step reaches children measurement results to fixed points, so the repeated calculations in `#4` step are elided.

components/DataTable/src/DataTable/DataRow.cs
```diff
@@ -48,180 +46,178 @@ private void DataRow_Unloaded(object sender, RoutedEventArgs e)
...
         }
 
-        return panel;
+        return _parentTable;
     }
 
     /// <inheritdoc/>
     protected override Size MeasureOverride(Size availableSize)
     {
+        //Debug.WriteLine($"DataRow.MeasureOverride");
```
Avid29, line 71: This should be removed, not just commented, before this PR can be accepted.

components/DataTable/src/DataTable/DataColumn.cs
```diff
@@ -10,21 +10,34 @@ namespace CommunityToolkit.WinUI.Controls;
...
+    /// NaN means that the column size is no yet calculated.
+    /// </summary>
+    internal double CurrentWidth { get; set; } = double.NaN;
+
     /// <summary>
-    /// Gets or sets the width of the largest child contained within the visible <see cref="DataRow"/>s of the <see cref="DataTable"/>.
+    /// Gets the internal calculated or manually set width of this column, as a positive value.
     /// </summary>
-    internal double MaxChildDesiredWidth { get; set; }
+    internal double ActualCurrentWidth => double.IsNaN(CurrentWidth) ? 0 : CurrentWidth;
```
Avid29, line 29: I'm not a fan of this. The implication that the current width can ever be negative seems like a design flaw.

components/DataTable/src/DataTable/DataTable.cs
```diff
@@ -47,120 +42,209 @@ public double ColumnSpacing
...
+        return new Size(totalWidth, maxHeight);
     }
 
     /// <inheritdoc/>
     protected override Size ArrangeOverride(Size finalSize)
     {
-        double fixedWidth = 0;
-        double proportionalUnits = 0;
-        double autoSized = 0;
+        //Debug.WriteLine($"DataTable.ArrangeOverride");
```
Avid29, line 185:
```suggestion
```

I'm going to convert this PR to a draft PR.

Please remove all debug artifacts before marking it ready for review

components/DataTable/src/DataTable/DataColumn.cs
```diff
@@ -10,21 +10,34 @@ namespace CommunityToolkit.WinUI.Controls;
...
+
+    internal bool IsAbsolute => DesiredWidth.IsAbsolute;
+
+    internal bool IsAuto => DesiredWidth.IsAuto;
+
+    internal bool IsStar => DesiredWidth.IsStar;
 
     /// <summary>
-    /// Gets or sets the internal copy of the <see cref="DesiredWidth"/> property to be used in calculations, this gets manipulated in Auto-Size mode.
+    /// Returns <see langword="true"/> if the column width is fixed with the manual adjustment.
```
Avid29, line 38:
```suggestion
    /// Gets whether or not the column width is fixed with the manual adjustment.
```

This is a more standard property comment format

Avid29 (commented):
I'd like to help get this control sorted out a bit. Here's some brief comments from what I saw here.

I'm also curious to know why there's so many open PRs the same control, by the same user, when the first seems like a pretty severe refactor. Seems like a recipe for frustrating merge conflicts.

components/DataTable/src/DataTable/DataColumn.cs
9rnsr: ? This value can never be negative; it will always be either `NaN` or a positive value.

PR #782, opened 2026-02-23
Remove support for case where the DataRow is descendant of a Grid or DataTable

9rnsr:
Separate PR from #781

The removed logic would have been useless or more harmful, because of the reasons:

1) `DataTable` control is designed to represent 'header' of a table, not for containing the list of `DataRow`s.
2) `Grid` control is often used to lay out other controls rather than constructing a table, so there was possibilities that completely unrelated `Grid`s would have been caught as a parent.


components/DataTable/src/DataTable/DataRow.cs
vgromfeld, line 48:
```suggestion
        // 1. Get parent ItemsPresenter to find header
```

vgromfeld (approved):

PR #783, opened 2026-02-23
Prevent to focus DataColumn itself by default

9rnsr:
Separate PR from #781

From the design concept, `DataColumn` control simply provides a space for the header content, so it should not interact with keyboard operations by default. If focusable something is actually needed, `DataTable` users can place any controls like a button there.


michael-hawker:
Cool, I see this in the current gallery where there's a ghost tab stop, will just double-check that the ContentSizer is still navigable (though I noticed that it's not working with keyboard for adjusting column size, may be specific to the DataTable as I see this working on the other Sizer scenarios).

9rnsr:
@michael-hawker After this PR change, the tab navigation still move the keyboard focus to the ContentSizer, and then Left and right keys can resize the column.

image

michael-hawker (approved):
Looks good. I found a case where the sample control itself is grabbing focus, so need to fix that. I also still see they keyboard issue, but found a fix for now (at least partial, the behavior isn't great, but it works). I'll open other issues to follow-up on. Thanks!

PR #784, opened 2026-02-23
Provide suitable DataRow behavior when DataTable is not present

9rnsr:
Separate PR from #781

This added behavior just covers a corner case. If a user fails to place corresponding `DataTable` at correct position, displaying somethings is better than nothing.

<del>Each rows layout their columns independent, because there's no main controller to remember the column widths, and then it looks like a horizontal `StackPanel`.</del>
If there's no main controller to remember the column widths, each DataRow uses the full width of its parent element to provide equal widths to its child elements.

### DIFF
--- a/components/DataTable/samples/DataRowWithoutDataTableSample.xaml
+++ b/components/DataTable/samples/DataRowWithoutDataTableSample.xaml
@@ -1,0 +1,38 @@
+<!--  Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information.  -->
+<Page x:Class="DataTableExperiment.Samples.DataRowWithoutDataTableSample"
+      xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+      xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+      xmlns:behaviors="using:CommunityToolkit.WinUI.Behaviors"
+      xmlns:controls="using:CommunityToolkit.WinUI.Controls"
+      xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+      xmlns:interactivity="using:Microsoft.Xaml.Interactivity"
+      xmlns:local="using:DataTableExperiment.Samples"
+      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+      mc:Ignorable="d">
+
+    <ListView ItemsSource="{x:Bind InventoryItems}">
+
+        <!-- no DataTable in <ListView.Header> -->
+
+        <ListView.ItemTemplate>
+            <DataTemplate x:DataType="local:InventoryItem">
+                <controls:DataRow>
+                    <TextBlock VerticalAlignment="Center"
+                               Text="{x:Bind Id}" />
+                    <TextBlock VerticalAlignment="Center"
+                               Text="{x:Bind Name}" />
+                    <TextBlock VerticalAlignment="Center"
+                               Text="{x:Bind Description}" />
+                    <TextBlock VerticalAlignment="Center"
+                               Text="{x:Bind Quantity}" />
+                </controls:DataRow>
+            </DataTemplate>
+        </ListView.ItemTemplate>
+        <!--<ListView.ItemContainerStyle>
+            <Style BasedOn="{StaticResource DefaultListViewItemStyle}"
+                   TargetType="ListViewItem">
+                <Setter Property="HorizontalContentAlignment" Value="Left" />
+            </Style>
+        </ListView.ItemContainerStyle>-->
+    </ListView>
+</Page>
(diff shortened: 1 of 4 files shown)


michael-hawker:
Was thinking about this scenario a bit more over the weekend, and I see a couple of directions:

1. If they just have a `DataRow` it's better to just have another panel there instead if they don't intend to use it as it's intended. Providing another behavior may be confusing or not lead them to understanding that there's more to be done. Probably something to think about if we can indicate this different state another way.
 
2. If we did want a default behavior. Probably doing a more `EqualPanel` type behavior where we space each item out based on the available width / number of items would make more sense, as then as long as the developer has the same number of items in each row (which they would with a data template) then it evenly spaces each item and maintains columns between the rows compared to them being jagged.

9rnsr:
The jagged layout of each row is exactly what I intended, because it can strongly recommend that the `DataRow` control is being used incorrectly.

…But, as you pointed out, there’s less real practical benefit to that behavior, and I can admit that your idea is more practical. I’ll update the default behavior so that each child element in the `DataRow` is given an equal width.

9rnsr:
Updated. To check the new `DataRow` behavior, I added a sample.

components/DataTable/src/DataTable/DataRow.cs
```diff
@@ -178,19 +194,43 @@ protected override Size MeasureOverride(Size availableSize)
...
     protected override Size ArrangeOverride(Size finalSize)
     {
+        // If we don't have a grid, provides equal widths to child elements.
+        if (_parentPanel is null)
+        {
+            if (Children.Count == 0)
+                return new Size(0, finalSize.Height);
+
+            double x = 0;
+            double width = finalSize.Width / Children.Count;
```
michael-hawker, line 204: We need the visible child count here otherwise we'll make everything smaller and not take up the full width, we can cache this from the measure step.

components/DataTable/src/DataTable/DataRow.cs
```diff
@@ -84,17 +84,33 @@ protected override Size MeasureOverride(Size availableSize)
 
         double maxHeight = 0;
 
-        if (Children.Count > 0)
+        // If we don't have a grid, provides equal widths to child elements.
+        if (_parentPanel is null)
```
michael-hawker, line 88: For both the measure and arrange bit we should probably abstract these to sub-methods to just make it a bit easier to contain.

components/DataTable/samples/DataRowWithoutDataTableSample.xaml
```diff
@@ -0,0 +1,38 @@
...
+      xmlns:interactivity="using:Microsoft.Xaml.Interactivity"
+      xmlns:local="using:DataTableExperiment.Samples"
+      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+      mc:Ignorable="d">
+
+    <ListView ItemsSource="{x:Bind InventoryItems}">
+
+        <!-- no DataTable in <ListView.Header> -->
+
+        <ListView.ItemTemplate>
```
michael-hawker, line 17: Sample is nice here thanks, this does seem like a useful base scenario, though someone could just use an EqualPanel here just as easily, it does extend itself later to just adding the rest of the pattern if needed.

components/DataTable/samples/DataRowWithoutDataTableSample.xaml
```diff
@@ -0,0 +1,38 @@
...
+                               Text="{x:Bind Quantity}" />
+                </controls:DataRow>
+            </DataTemplate>
+        </ListView.ItemTemplate>
+        <!--<ListView.ItemContainerStyle>
+            <Style BasedOn="{StaticResource DefaultListViewItemStyle}"
+                   TargetType="ListViewItem">
+                <Setter Property="HorizontalContentAlignment" Value="Left" />
+            </Style>
+        </ListView.ItemContainerStyle>-->
```
michael-hawker: Unneeded?

michael-hawker (commented):
Few small comments. @niels9001 any thoughts on this as a default partial behavior?

components/DataTable/samples/DataRowWithoutDataTableSample.xaml
9rnsr: Yes. I deleted these lines.

components/DataTable/src/DataTable/DataRow.cs
9rnsr: I'm not sure we need to add one more integer value in each `DataRow` that is rarely used...

components/DataTable/src/DataTable/DataRow.cs
9rnsr: Yes, we can refactor the code after the `DataTable` improvements finished.

components/DataTable/samples/DataRowWithoutDataTableSample.xaml
9rnsr: Remember that it is just an explanation sample for the default behavior. For example, there is no way to specify spacing between columns.

Issue #611, opened 2024-11-25
[MarkdownTextBlock] Turn everything in Config into DependencyProperty


nerocui:
### Describe the bug

Current implementation requires setting a Config object in either code-behind or view model. We should do that behind the scene and use DependencyProperty to forward all the settings from the consumer of the control to the internal config object.

### Steps to reproduce

```text
Use the control without the config and the control will crash.

<controls:MarkdownTextBlock Grid.Column="0"
                            Config="{x:Bind LiveMarkdownConfig, Mode=OneTime}"
                            Text="{x:Bind MarkdownTextBox.Text, Mode=OneWay}" />
```

### Expected behavior

The control should work fine without any property being set. The developer should be able to set any property of config by using a dependency property instead of contracting the config in code.

### Code Platform

- [x] UWP
- [x] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [ ] Windows 11 21H2 (Build 22000)
- [ ] Other (specify)

### App minimum and target SDK version

- [ ] Windows 10, version 1809 (Build 17763)
- [ ] Windows 10, version 1903 (Build 18362)
- [ ] Windows 10, version 1909 (Build 18363)
- [ ] Windows 10, version 2004 (Build 19041)
- [ ] Other (specify)

### Help us help you

Yes, I'd like to be assigned to work on this item.

michael-hawker:
The default config should be there by default too, so if you don't do anything the control "just works"™️ out-of-the-box too. Saw this hit when trying to migrate from the old version.

JimmyRespawn:
It'll be nice to use it without setting the config. In a situation like using it in Listview, MarkdownTextBlock is much easier to setup in v7.x. Also it would be nice to have OnImageClick return in v8 because it's easier call in-box image viewer to see images inside markdown.

PR #785, opened 2026-02-24
 MarkdownTextBlock: ThemeResource support, DP migration & performance fixes

niels9001:
# MarkdownTextBlock: ThemeResource support, DP migration & performance fixes

## Summary

This PR modernizes the `MarkdownTextBlock` control's theming architecture and fixes several performance/memory issues. The control now supports proper `{ThemeResource}` brushes that update automatically on Light/Dark/HighContrast theme switches — previously, theme changes had no effect on markdown styling.

**Addresses**: #611

---

### Theming & DependencyProperty migration

Previously, all styling lived in plain C# classes (`MarkdownThemes` with auto-properties, wrapped by `MarkdownConfig`). This meant `{ThemeResource}` couldn't target them, and theme switches (Light ↔ Dark) had no effect on markdown brush colors.

**New architecture**: All 67+ styling properties are now **DependencyProperties directly on `MarkdownTextBlock`**. The intermediary `MarkdownThemes` and `MarkdownConfig` classes are deleted entirely. TextElement renderers (`MyHeading`, `MyCodeBlock`, `MyQuote`, `MyTable`, etc.) receive the `MarkdownTextBlock` control reference and read DPs directly from it at render time.

```xml
<controls:MarkdownTextBlock
    H1Foreground="{ThemeResource MyCustomBrush}"
    CodeBlockBackground="{StaticResource CodeBg}"
    InlineCodeCornerRadius="4" />
```

#### DPs added to `MarkdownTextBlock.Properties.cs`

| Category | Properties |
|----------|-----------|
| **Headings** (18) | `H1`–`H6` × `FontSize`, `Foreground`, `FontWeight`, `Margin` |
| **Inline Code** (7) | `InlineCodeBackground`, `InlineCodeForeground`, `InlineCodeBorderBrush`, `InlineCodeBorderThickness`, `InlineCodeCornerRadius`, `InlineCodePadding`, `InlineCodeFontSize`, `InlineCodeFontWeight` |
| **Code Blocks** (9) | `CodeBlockBackground`, `CodeBlockForeground`, `CodeBlockBorderBrush`, `CodeBlockBorderThickness`, `CodeBlockPadding`, `CodeBlockMargin`, `CodeBlockFontFamily`, `CodeBlockCornerRadius` |
| **Quotes** (8) | `QuoteBackground`, `QuoteForeground`, `QuoteBorderBrush`, `QuoteBorderThickness`, `QuoteMargin`, `QuotePadding`, `QuoteBarMargin`, `QuoteCornerRadius` |
| **Tables** (6) | `TableHeadingBackground`, `TableBorderBrush`, `TableBorderThickness`, `TableCellPadding`, `TableMargin`, `TableCornerRadius` |
| **Horizontal Rule** (4) | `HorizontalRuleBrush`, `HorizontalRuleThickness`, `HorizontalRuleMargin` |
| **Links** (1) | `LinkForeground` |
| **Images** (3) | `ImageMaxWidth`, `ImageMaxHeight`, `ImageStretch` |
| **Paragraphs/Lists** (4) | `ParagraphMargin`, `ParagraphLineHeight`, `ListBulletSpacing`, `ListGutterWidth` |
| **Bold** (1) | `BoldFontWeight` |
| **Config** (3) | `BaseUrl`, `ImageProvider`, `SVGRenderer` (previously on `MarkdownConfig`) |

#### Default style (`MarkdownTextBlock.xaml`)

- **`ThemeDictionaries`** with `Default`, `Light`, and `HighContrast` dictionaries define all brush resources (e.g., `MarkdownTextBlockH1Foreground` → `TextFillColorPrimaryBrush`). HighContrast uses `SystemColorButtonTextColorBrush`, `SystemColorHotlightColorBrush`, etc.
- **Non-brush values** (font sizes, margins, thicknesses, corner radii, font weights) are set as literal values in style setters — not as named resources.
- Follows the same pattern as `SettingsControls` in this repo.

#### Rendering architecture changes

- **`WinUIRenderer`** — Removed `Config` property. Constructor now takes `(MyFlowDocument, MarkdownTextBlock)`. Already had a `MarkdownTextBlock` property.
- **All 11 TextElement classes** — Constructors changed from `MarkdownThemes`/`MarkdownConfig` parameter to `MarkdownTextBlock`. Each reads DPs directly (e.g., `_control.H1FontSize`, `_control.CodeBlockBackground`).
- **All 8 ObjectRenderers + HtmlWriter** — Pass `renderer.MarkdownTextBlock` instead of `renderer.Config.Themes`.
- **Batched re-render** — All theme DPs share a single `OnThemePropertyChanged` callback that uses `DispatcherQueue.TryEnqueue` with a boolean guard to coalesce multiple simultaneous DP changes (e.g., 20+ brush updates on theme switch) into one re-render.
- **`ActualThemeChanged`** handler as belt-and-suspenders for theme switching, using the same batching guard.

### Performance & memory fixes

1. **Fixed `HttpClient` socket leak** (`MyImage.cs`)  
   Each image load created `new HttpClient()`, never disposed. `HttpClient` maintains connection pools internally, so per-instance creation leaks socket handles. Replaced with a `static readonly HttpClient` shared across all image loads.

2. **Fixed `Image.Loaded` event handler leak** (`MyImage.cs`)  
   `_image.Loaded += LoadImage` was never unsubscribed. On every theme-change re-render, the old `MyImage` objects remained rooted via the event handler delegate, preventing garbage collection. Now unsubscribes immediately at the top of the handler.

3. **Fixed `ActualThemeChanged` event leak** (`MarkdownTextBlock.xaml.cs`)  
   The event was subscribed in the constructor but never unsubscribed. Moved to `Loaded`/`Unloaded` lifecycle handlers, matching the pattern used by the old CommunityToolkit MarkdownTextBlock control.

4. **Eliminated per-image `DefaultSVGRenderer` allocation** (`MyImage.cs`)  
   Every image without a custom SVG renderer created `new DefaultSVGRenderer()`. Since it's stateless, replaced with a `static readonly` singleton.

5. **Fixed double-scan on `HtmlNodeCollection`** (`HtmlWriter.cs`)  
   `node.ChildNodes.Remove(node.ChildNodes.FirstOrDefault(...))` scanned the collection twice — once to find the node, once to remove it. Refactored to a single `FirstOrDefault` + null-check + `Remove`.

6. **Improved text buffer growth strategy** (`WinUIRenderer.cs`)  
   When text exceeded the 1024-char buffer, the entire buffer was replaced with an exact-fit `ToCharArray()` — likely too small for the next call, causing repeated reallocations. Now uses a doubling strategy (`Math.Max(length, buffer.Length * 2)`).

### Sample updates

- **Example sample** simplified — no longer needs `Config` binding; defaults come from the XAML style automatically.
- **Custom Theme sample** updated to bind directly to control DPs via `x:Bind` function bindings.
- **"Apply Changes" button removed** from theme options — changes are now live via bindings.

---

## Breaking changes

- `MarkdownConfig` and `MarkdownThemes` classes are **deleted**. Users who were setting `Config` should set properties directly on the control:
  ```diff
  - markdown.Config = new MarkdownConfig { Themes = new MarkdownThemes { H1FontSize = 32 } };
  + markdown.H1FontSize = 32;
  ```
- `BaseUrl`, `ImageProvider`, `SVGRenderer` are now DPs on `MarkdownTextBlock` (previously on `MarkdownConfig`):
  ```diff
  - markdown.Config = new MarkdownConfig { ImageProvider = myProvider };
  + markdown.ImageProvider = myProvider;
  ```

## Testing

- XAML compiler: zero errors across all target frameworks
- Existing `ImageProviderConstraintTest` updated to use new DP API
- Theme switching verified: Light ↔ Dark ↔ HighContrast all update brushes correctly


niels9001:
@Arlodotexe @michael-hawker can we get this reviewed? We are still blocked by the textblocks not changing colors..

Lorenz5600:
Yes, a review would be great. We're also blocked by unreadable hint headers.

niels9001:
@Arlodotexe this PR has been pending for 4 months now.. When can we expect a review?

Arlodotexe:
> @Arlodotexe this PR has been pending for 4 months now.. When can we expect a review?

Thanks for bumping @niels9001. Organizing has been chaotic, especially with Michael leaving. I'm going through and reviewing ALL PRs with deliberate priority orientation-- this was on the list, but thanks to your ping and existing priority criteria it's now second to top of my agenda now (just behind repo-wide CI fixes).

I'll be attending this until we've closed it. Expect a review by today or tomorrow.

components/MarkdownTextBlock/src/MarkdownTextBlock.Properties.cs
```diff
@@ -125,10 +155,467 @@ private static void OnIsTextSelectionEnabledChanged(DependencyObject d, Dependen
         }
     }
 
-    public MarkdownConfig Config
+    // ── Theme DPs ───────────────────────────────────────────────────────
+    // Defaults come from the XAML style. C# defaults are type-appropriate
+    // fallbacks only. The shared property-changed callback batches multiple
+    // simultaneous DP updates (e.g. theme switch) into a single re-render.
+
+    private static void OnThemePropertyChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
```
Arlodotexe, line 172: - This seems to be the whole "shared callback batches multiple simultaneous updates" setup. 
    - I'm not convinced that this is doing what we expect it to.
    - Let me logically walk through the code:
        - Style change happens
        - Multiple properties are changed at the same time
        - DP callbacks don't fire perfectly concurrently
        - `OnThemePropertyChanged` is called multiple times, overlapping and sequentially
        - The very first entry sets `self._themePropertyChangeQueued = true;`
        - The second/third/etc method entry is dropped due to the `if` gate on `_themePropertyChangeQueued` needing to be false
        - The first method entry continues on the DispatcherQueue and unsets `_themePropertyChangeQueued` to `false`
        - `ApplyText` is called on the dispatcher queue, still caused by the first entry.
    - This looks like a clear issue to me:
        - Since the first method entry is the only one that can call `ApplyText`, it may finish before *all* DP updates can happen if DP callbacks aren't perfectly concurrent, meaning it may also call `ApplyText` multiple times while flipping `_themePropertyChangeQueued` back and forth in both concurrent and sequential update scenarios. 
        - That means one property change will always work, two property changes may even work concurrently, but the odds of duplicating the `ApplyText` call may go up the more property changes are made via a single style swap. 
    - I'll want to flag this, but alternatives are an open question:
        - Should we be using a debounce?
        - Should we find another way to wait until all DP callbacks are done firing?
    - If the intent is to batch updates rather than prevent duplicate/concurrent `ApplyText` calls, then either way `ApplyText` should be handled by the *final* DP callback entry, not the first one. 
    - If the intent is to prevent duplicate/concurrent `ApplyText` calls and not to batch together rapid-fire DP updates, the comment should be updated to say that clearly but the code will work as-is.

components/MarkdownTextBlock/src/MarkdownTextBlock.xaml
```diff
@@ -1,12 +1,238 @@
-﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+﻿<!--  Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information.  -->
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
                     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
                     xmlns:controls="using:CommunityToolkit.WinUI.Controls">
 
+    <ResourceDictionary.ThemeDictionaries>
+        <ResourceDictionary x:Key="Default">
```
Arlodotexe, line 7: "Default" is ambiguous about whether it means Dark or Light-- depends on application and system settings. It'll cause problems when switching themes.

[Docs here](https://learn.microsoft.com/en-us/windows/apps/develop/platform/xaml/xaml-theme-resources#troubleshooting-theme-resources) explicitly call this out and recommend against it, with a full explainer as to why.

components/MarkdownTextBlock/src/MarkdownTextBlock.xaml
```diff
@@ -1,12 +1,238 @@
...
+        <Setter Property="QuoteBorderBrush" Value="{ThemeResource MarkdownTextBlockQuoteBorderBrush}" />
+        <Setter Property="QuoteForeground" Value="{ThemeResource MarkdownTextBlockQuoteForeground}" />
+        <Setter Property="QuoteBorderThickness" Value="4,0,0,0" />
+        <Setter Property="QuoteMargin" Value="0,4,0,4" />
+        <Setter Property="QuotePadding" Value="4" />
+        <Setter Property="QuoteCornerRadius" Value="4" />
+        <Setter Property="QuoteBarMargin" Value="0,0,4,0" />
+
+        <!--  Image properties  -->
+        <Setter Property="ImageMaxWidth" Value="0" />
```
Arlodotexe, line 224: Is it intentional that `ImageMaxWidth` and `ImageMaxHeight` are both `0` by default?

components/MarkdownTextBlock/src/MarkdownTextBlock.xaml.cs
```diff
@@ -54,6 +53,31 @@ public MarkdownTextBlock()
...
+    }
+
+    private void OnUnloaded(object sender, RoutedEventArgs e)
+    {
+        this.ActualThemeChanged -= OnActualThemeChanged;
+    }
+
+    private void OnActualThemeChanged(FrameworkElement sender, object args)
+    {
+        if (!_themePropertyChangeQueued)
```
Arlodotexe, line 75: Same method shape and concern as flagged in [this comment](https://github.com/CommunityToolkit/Labs-Windows/pull/785#discussion_r3424740478)

components/MarkdownTextBlock/src/MarkdownTextBlock.xaml.cs
```diff
@@ -107,40 +123,38 @@ private void ApplyText(bool rerender)
...
+            if (!DisableLinks) _renderer.ObjectRenderers.Add(new ContainerInlineRenderer());
+
+            // Extension renderers
+            if (UsePipeTables) _renderer.ObjectRenderers.Add(new TableRenderer());
+            if (UseTaskLists) _renderer.ObjectRenderers.Add(new TaskListRenderer());
+            if (!DisableHtml) _renderer.ObjectRenderers.Add(new HtmlInlineRenderer());
         }
+        _pipeline.Setup(_renderer);
+
+        ApplyText(true);
```
Arlodotexe: Looking [here](https://github.com/CommunityToolkit/Labs-Windows/blob/b7e6d87655e3e9cfd75047aa8525389210547aca/components/MarkdownTextBlock/src/MarkdownTextBlock.xaml.cs#L106), the bool param here is called `rerender`. Internally, this method body uses the param to gate whether `_renderer.ReloadDocument` is called. 

It's unclear from the code alone why this param value was changed from `false` to `true` in this method for this PR. 

@niels9001 Any insight you can provide on this? If the rationale can't be found, we should probably expose this param through the containing `Build` method so it can be deliberately tuned by the needs of each caller. If it can be changed arbitrarily like this (unless it was deliberate but not obvious), then it may be underspecified throughout the codebase.

components/MarkdownTextBlock/src/Renderers/WinUIRenderer.cs
```diff
@@ -125,14 +118,10 @@ public void WriteText(string? text, int offset, int length)
...
             if (length > _buffer.Length)
             {
-                _buffer = text.ToCharArray();
-                WriteText(new string(_buffer, offset, length));
-            }
-            else
-            {
-                text.CopyTo(offset, _buffer, 0, length);
-                WriteText(new string(_buffer, 0, length));
+                _buffer = new char[Math.Max(length, _buffer.Length * 2)];
```
Arlodotexe: - Why exactly *2 the buffer size? Thinking it through inline:
    - Functionally still intended to expand the buffer
        - but now takes twice the buffer size
            - only when twice the buffer size is bigger than the given length
        - combined with the `if` gate, this means it always doubles the buffer if it's too small to fit the requested length.
    - Intent for the change appears to be avoiding allocations coming from `text.ToCharArray();`
    - I'm not sure that this is the best way to do this. 
    - Potential issues:
        - If the requested length is more than twice the size of the given buffer, blindly doubling the buffer won't expand enough to fit the text copied into it-- forces Max.Math fallback.
        - The Math.Max choosing between `length`...
            - redundant, or patching the x2 issue...? 
            - `length` is functionally ambiguous here:
                - if `length` is raw byte count, expanding the buffer to match it is fine
                    - comparison of `if (length > _buffer.Length)` implies this is true
                - if `length` is literal character count, this isn't a functionally equivalent code change to what we had before.
            - Assuming `length` means byte count and not character count, then choosing to Math.Max between it and doubling the buffer size may be redundant-- `length` should be directly usable unless a bad value is being provided upstream.
        - Overall, this change has a bit of code smell that needs investigated and corrected.
        - We should revert this and flag/track for deferral/follow-up in another issue/PR so the ready DP adjustments can make it through.

components/MarkdownTextBlock/src/TextElements/MyFlowDocument.cs
```diff
@@ -14,38 +14,46 @@
 
 namespace CommunityToolkit.WinUI.Controls.TextElements;
 
+/// <summary>
+/// Represents a flow document that wraps a <see cref="Microsoft.UI.Xaml.Controls.RichTextBlock"/> for rendering markdown or HTML content.
```
Arlodotexe: On MUX: RichEditBox exists in WinUI 2/MUX, but RichTextBlock only exists in WinUI 3/MUX while the UWP version of RichTextBox comes from the platform under WUX, so the namespace here needs adjusted slightly.

Our globalusings should auto-include MUXC on WASDK and WUXC on UWP out of the box, so simply shortening it should work. 

```suggestion
/// Represents a flow document that wraps a <see cref="RichTextBlock"/> for rendering markdown or HTML content.
```

components/MarkdownTextBlock/src/TextElements/MyFlowDocument.cs
```diff
@@ -14,38 +14,46 @@
...
     private HtmlNode? _htmlNode;
     private RichTextBlock _richTextBlock = new RichTextBlock();
     private MarkdownObject? _markdownObject;
 
-    // useless property
+    /// <summary>Gets or sets the text element (unused, required by <see cref="IAddChild"/>).</summary>
     public TextElement TextElement { get; set; } = new Run();
-    //
 
+    /// <summary>Gets or sets the underlying <see cref="Microsoft.UI.Xaml.Controls.RichTextBlock"/>.</summary>
```
Arlodotexe: Same comment as [this one](https://github.com/CommunityToolkit/Labs-Windows/pull/785#discussion_r3425604604) regarding MUX vs WUX on RichTextBlock:

```suggestion
    /// <summary>Gets or sets the underlying <see cref="RichTextBlock"/>.</summary>
```

components/MarkdownTextBlock/src/TextElements/MyImage.cs
```diff
@@ -120,10 +125,7 @@ private async void LoadImage(object sender, RoutedEventArgs e)
             }
             else
             {
-                HttpClient client = new HttpClient();
-
-                // Download data from URL
-                HttpResponseMessage response = await client.GetAsync(_uri);
+                HttpResponseMessage response = await _sharedHttpClient.GetAsync(_uri);
```
Arlodotexe, line 126: This should be exposed as a property that can be set on the `MarkdownTextBlock` itself so it can be changed by the code consumer and shared everywhere (not just once per rendered image).

Not a blocker to closing this PR; it's an improvement over what we had but it still leaves room for further improvement.

components/MarkdownTextBlock/src/TextElements/MyImage.cs
```diff
@@ -89,6 +93,7 @@ private void Init()
 
     private async void LoadImage(object sender, RoutedEventArgs e)
     {
+        _image.Loaded -= LoadImage;
```
Arlodotexe, line 94: - Reading as: Immediately unsubs the `_image.Loaded -= LoadImage;` event, making the handler a one-time use (unless resubscribed at end of this method body)
- I'm not seeing the re-sub to the `_image.Loaded` event.
    - This changes the behavior, specifically making each `MyImage` instance single-use. 
    - Once an image is loaded, it cannot be re-loaded, even if the `MyImage` instance is removed and re-added to the visual tree. 
    - If the intent here is to prevent the event body from running more than once at a time (semaphore-like), we should be resubscribing the event at the end of the method body.

components/MarkdownTextBlock/src/TextElements/MyImage.cs
```diff
@@ -48,16 +52,16 @@ public MyImage(LinkInline linkInline, Uri uri, MarkdownConfig config)
     }
 
 #pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
-    public MyImage(HtmlNode htmlNode, MarkdownConfig? config)
+    public MyImage(HtmlNode htmlNode, MarkdownTextBlock? control)
```
Arlodotexe: This doesn't need to be nullable-- the null suppression in the body was a red flag that helped catch it, but I verified locally and the only reference to this ctor in our own code is from `HtmlWriter.cs` L69 where the provided reference `renderer.MarkdownTextBlock` is never null. 

This means the nullability is an artifact of the prior `MarkdownConfig? config` being nullable. I can't think of any reason this control needs to operate without a `MarkdownTextBlock` instance attached, so let's align it with the other TextElement controls and make it non-nullable. 

It should clean things up here nicely.

Arlodotexe (changes requested):
Great PR @niels9001, many much-needed quality improvements here! 

Reviewed with a fine-tooth comb, took me approx. ~6h 15m from first source file to all changed files reviewed.

11 review comments added. Some to address or simply clarify before closing, some to defer to later.

Reviewed in order of source, then samples, then tests. All other changes not commented on were validated for internal consistency and overall correctness.

To top it off, I've visually validated via a local temp sample on both WASDK and UWP that all new style assignments and DPs were referencing valid and usable brushes in both Light/Dark themes at sample-control-level `ThemeResource` swap (not system, see "Default" vs "Dark" review comments):

Image 

Image

Human/AI attribution & disclosure:
  - All file-by-file review and validation (structural, functional, design, performance) was done 100% by human hand based on:
    - Direct diff reading
    - MSLearn docs checks
     - Personal and [community](https://discord.com/channels/372137812037730304/580484470877061120/1516620980942929931)-based knowledge sourcing
      - AI-delegated local/repro validation (see below).
  - AI agent assistance was used:
      - To scaffold/verify throwaway UWP and WinUI 3 `<StaticResource ...>` repros.
      - At the end to surface and attend to deferred review items from the area/task log.
      - To scaffold the focused theme-resource coverage sample (screenshots above)

Full verbatim area/task log available on request.

components/MarkdownTextBlock/src/MarkdownTextBlock.xaml
niels9001: Copilot you are drunk: https://github.com/search?q=repo%3Amicrosoft%2Fmicrosoft-ui-xaml+x%3AKey%3D%22Default%22&type=code

components/MarkdownTextBlock/src/MarkdownTextBlock.Properties.cs
niels9001: Updated the comment to state the intent precisely. This is coalescing, not just dedupe: on a theme/style swap the DP callbacks fire **synchronously** on the UI thread, so the first update queues a single re-render on the dispatcher and sets `_themePropertyChangeQueued`; the rest of the burst is skipped by the flag. Because the queued callback only runs after the current synchronous work drains, `ApplyText` runs exactly once, after every property has been applied. The flag is shared with `OnActualThemeChanged` so the theme event and the DP updates it triggers collapse into the same re-render. Clarified in dae6b7c.

components/MarkdownTextBlock/src/MarkdownTextBlock.xaml.cs
niels9001: Same mechanism as [the OnThemePropertyChanged thread](https://github.com/CommunityToolkit/Labs-Windows/pull/785#discussion_r3424740478) — shared flag, synchronous DP callbacks, single queued re-render. Comment clarified in dae6b7c.

components/MarkdownTextBlock/src/MarkdownTextBlock.xaml.cs
niels9001: Good catch. Originally `Build()` ran `ApplyText(false)` because the renderer was created fresh in the same call, so there was nothing to reload. After the refactor `Build()` is the single entry from `OnApplyTemplate` and the renderer is cached, so a reload gives a clean re-render if the template is (re)applied. To make this deliberately tunable rather than hard-coded, I exposed it as `Build(bool rerender = true)` and `OnApplyTemplate` calls the default. Done in dae6b7c.

components/MarkdownTextBlock/src/Renderers/WinUIRenderer.cs
niels9001: Reverted to the pre-PR `WriteText` buffer implementation in dae6b7c so the DP work can land cleanly. I'll track the allocation optimization (avoiding `ToCharArray`) as a separate follow-up.

components/MarkdownTextBlock/src/MarkdownTextBlock.xaml
niels9001: Intentional — `0` means 'no max constraint', and the consuming code gates on `> 0` (see `MyImage.LoadImage`), so the image falls back to its natural / markdown-specified size. Added a clarifying comment on the setters in dae6b7c.

components/MarkdownTextBlock/src/TextElements/MyFlowDocument.cs
niels9001: Done in dae6b7c — shortened to `<see cref=\"RichTextBlock\"/>` so global usings resolve MUXC/WUXC per platform.

components/MarkdownTextBlock/src/TextElements/MyFlowDocument.cs
niels9001: Done in dae6b7c.

components/MarkdownTextBlock/src/TextElements/MyImage.cs
niels9001: Done — made the `MarkdownTextBlock` parameter non-nullable and removed the related null suppression in dae6b7c.

components/MarkdownTextBlock/src/TextElements/MyImage.cs
niels9001: Agreed, not a blocker. I'll track exposing a shared `HttpClient` on `MarkdownTextBlock` as a follow-up improvement.

components/MarkdownTextBlock/src/TextElements/MyImage.cs
niels9001: This is intentionally single-use per instance: `MyImage` instances aren't reused — the document is fully rebuilt (new `MyImage` instances) on every re-render — so a re-add never needs to reload the same instance. The early unsubscribe plus the `_loaded` guard ensures the async load body runs at most once. I'll leave behavior as-is; happy to revisit if you'd prefer an explicit re-subscribe model.

PR #786, opened 2026-02-24
Add internal methods in DataColumn

9rnsr:
Separate PR from #781

Currently there are three `DataColumn` types for the layout logic. I'll add two more types for later improvement.

### DIFF
--- a/components/DataTable/src/DataTable/DataColumn.cs
+++ b/components/DataTable/src/DataTable/DataColumn.cs
@@ -16,6 +16,8 @@ public partial class DataColumn : ContentControl
 
     private WeakReference<DataTable>? _parent;
 
+    internal DataTable? DataTable => _parent?.TryGetTarget(out DataTable? parent) == true ? parent : null;
+
     /// <summary>
     /// Gets or sets the width of the largest child contained within the visible <see cref="DataRow"/>s of the <see cref="DataTable"/>.
     /// </summary>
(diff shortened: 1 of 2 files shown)


components/DataTable/src/DataTable/DataColumn.cs
```diff
@@ -16,6 +16,8 @@ public partial class DataColumn : ContentControl
 
     private WeakReference<DataTable>? _parent;
 
+    internal DataTable? DataTable => _parent?.TryGetTarget(out DataTable? parent) == true ? parent : null;
```
michael-hawker:
```suggestion
    //// Pass-thru convenience helper to get reference from WeakReference; computed each time, do not store result!
    private DataTable? DataTable => _parent?.TryGetTarget(out DataTable? parent) == true ? parent : null;
```

Does this need to be internal? I'm a little leery of exposing the reference. I was trying to see where in #781 it was used, but it's been a bit hard to follow.

Should the other properties below be public though to help others understand the current mode of the control for IsAbsolute, IsAuto, and IsStar.

9rnsr: You’re absolutely right. Since this `DataTable` property is only used internally within the `DataColumn` class, it can be made private. I’ll update it accordingly.

As for `IsAbsolute`, `IsAuto`, and `IsStar`, making those properties public would certainly make the current mode of the control easier to understand. However, doing so would also mean adding new public interfaces, which I wanted to avoid.

vgromfeld (approved):

PR #793, opened 2026-04-14
Fix escaping string literals

Sergio0694:
### Closes #788


Arlodotexe (approved):
Reviewed and approved, easy fix and great test coverage.

PR #794, opened 2026-04-22
Fix optimized DP setter codegen for 'string' properties

Sergio0694:
Follow-up to #792.

The optimized codegen for generated dependency property setters introduced in #792 calls `XamlBindingHelper.SetProperty*` directly, but it turns out that `XamlBindingHelper.SetPropertyFromString` does not work correctly when the value is `null` or an empty string. To work around this, the generator now special-cases `string` properties and emits this code instead of an unconditional helper call:

```csharp
if (value is null || value.Length == 0)
{
    SetValue(TestProperty, value);
}
else
{
    XamlBindingHelper.SetPropertyFromString(this, TestProperty, value);
}
```

This is applied to both the local-caching and no-caching setter paths.

The existing unit tests covering `string` and `string?` dependency properties (including the parameterized rows in `SingleProperty_MultipleTypes_WithNoCaching_DefaultValueIsOptimized` and `SingleProperty_WithCustomMetadataType_WithNoCaching`, plus the explicit `SingleProperty_String_WithLocalCache` test and others) have been updated to reflect the new generated code, so coverage for both setter paths is preserved.

### DIFF
--- a/components/DependencyPropertyGenerator/CommunityToolkit.DependencyPropertyGenerator.SourceGenerators/DependencyPropertyGenerator.Execute.cs
+++ b/components/DependencyPropertyGenerator/CommunityToolkit.DependencyPropertyGenerator.SourceGenerators/DependencyPropertyGenerator.Execute.cs
@@ -764,9 +764,29 @@ partial class DependencyPropertyGenerator
                                 field = value;
                             """, isMultiline: true);
 
-                        // If an optimized 'XamlBindingHelper' method is available, use it directly
-                        if (propertyInfo.XamlBindingHelperSetMethodName is string setMethodName)
+                        // If the property is of type 'string', we need a special path. That is because 'XamlBindingHelper.SetPropertyFromString'
+                        // doesn't work correctly for 'null' or empty strings, so we need to fall back to 'SetValue' in those cases.
+                        if (propertyInfo.TypeName == "string")
                         {
+                            writer.Write($$"""
+
+                                    if (value is null || value.Length == 0)
+                                    {
+                                        SetValue({{propertyInfo.PropertyName}}Property, value);
+                                    }
+                                    else
+                                    {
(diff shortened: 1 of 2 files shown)


components/DependencyPropertyGenerator/CommunityToolkit.DependencyPropertyGenerator.SourceGenerators/DependencyPropertyGenerator.Execute.cs
```diff
@@ -764,9 +764,29 @@ static string GetExpressionWithTrailingSpace(Accessibility accessibility)
...
 
-                        // If an optimized 'XamlBindingHelper' method is available, use it directly
-                        if (propertyInfo.XamlBindingHelperSetMethodName is string setMethodName)
+                        // If the property is of type 'string', we need a special path. That is because 'XamlBindingHelper.SetPropertyFromString'
+                        // doesn't work correctly for 'null' or empty strings, so we need to fall back to 'SetValue' in those cases.
+                        if (propertyInfo.TypeName == "string")
                         {
+                            writer.Write($$"""
+
+                                    if (value is null || value.Length == 0)
```
michael-hawker, line 773:
```suggestion
                                    if (string.IsNullOrEmpty(value))
```
QQ: I know it's equivalent but would using the string helper be clearer/more efficient in anyway? (i.e. would it be optimized by the compiler anyway vs. the function overhead?)

Sergio0694: No the method is literally just this code internally. If anything this avoids the need for inlining 🙂

Arlodotexe (approved):
Source change looks good, test coverage looks good. Cloned and ran the tests locally in VS, all pass. Approved!

PR #797, opened 2026-04-28
Fix for DataColumn to resize with keyboard

michael-hawker:
See issue https://github.com/CommunityToolkit/Windows/issues/849 from the Toolkit where it's hard for us to respond here, we need a bit of thinking about how to best respond here.


michael-hawker:
Oh, interesting, was just pulling #781 and that doesn't have this same issue... though I'm trying to understand why now... 🤔

michael-hawker:
Ah, I think there's more measure/arrange invalidation in 781 which is happening which just forces the issue. Would be curious to see if we could measure the measure passes, as minimizing those in the core scenarios would be great. Resizing the column here is an exception. Though in general, we're kind of fighting layout a bit with the design of this control, but the goal was to minimize those passes as we just want to be responding to the bare minimum of events:
- Column resized by drag
- Content resized in Auto column and now larger
- Control itself resized

Issue #796, opened 2026-04-28
[Gradient Slider] Gradient Stop Collection does not respond to external add/removals


Avid29:
The `GradientStop` has an event for the color updating and the offset updating, and the `GradientSliderThumb` will use them to update.

However, the `GradientStopCollection` does not appear to have any event for a stop being added or removed, so the `GradientSlider` is not subscribed to any event which handles the adding or removing of a stop.

PR #798, opened 2026-04-29
Added public explicit refresh method to GradientSlider

Avid29:
This PR introduces a rather brute-force "solution" to issue #796.

A new method `Refresh` is added to the `GradientSlider` which allows the developer to inform the control that the `GradientStopCollection` has been modified externally.

### DIFF
--- a/components/GradientSlider/src/GradientSlider.cs
+++ b/components/GradientSlider/src/GradientSlider.cs
@@ -57,6 +57,16 @@ public partial class GradientSlider : Control
         ];
     }
 
+    /// <summary>
+    /// Explicitly reloads the thumbs for the current <see cref="GradientStopCollection"/>.
+    /// </summary>
+    /// <remarks>
+    /// This method must be used if a <see cref="GradientStop"/> is added or removed from the collection externally.
+    /// It need not be used when a <see cref="GradientStop.Color"/> or <see cref="GradientStop.Offset"/> is modified, as those changes are automatically observed.
+    /// See <see href="https://github.com/CommunityToolkit/Labs-Windows/issues/796"/> for more details.
+    /// </remarks>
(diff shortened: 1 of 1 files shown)


michael-hawker:
@Avid29 if you make this an `ObservableCollection` does that change anything? That's the crux of the problem. I don't remember if that always will work with how the XAML compiler initializes the collections in all scenarios though. Especially if the dev does a GradientStopCollection in code-behind and assigns it later.

https://github.com/CommunityToolkit/Labs-Windows/blob/f437e525e57d33c56315ae68577d12d80d21dfb0/components/GradientSlider/src/GradientSlider.cs#L45-L57

Would it be nicer to just have an `AddStop` and `RemoveStop` method on the control to have it be a passthru that can then have it update?

Avid29:
@michael-hawker I'm not sure I understand what you mean. That code is using collection initialization syntax to initialize a `GradientStopCollection`, which is its own concrete type and the type for the main input/output property of the control. It cannot be simply replaced with an `ObservableCollection`.

However, whether or not using `Microsoft.UI.Xaml.Media.GradientStopCollection` for the main data property is right to use at all is a valuable discussion. Here's a pros and cons list off the top of my head:

| Pros | Cons |
| ----- | ----- |
| Is a pre-existing type |  Does not have a collection updated event |
| Can bind the control to other GradientBrushes in the Visual Tree | Uses `Microsoft.UI.**` type for data (which is weird to put in the ViewModel for MVVM) |

Using a `Microsoft.UI.**` type for data bothers me particularly. I like to fully separate my View, ViewModel, and Model into separate projects where only the View has any UI package references. Currently there's no solution which makes that feasible.

I think my perspective is to leave as is (with this change) for now, and look at something using a `System.Numerics.**` color when we get to .NET 11.

PR #799, opened 2026-04-30
Add Uno support for DataTable

### LINKED ISSUE #559: DataTable issues on Uno
## Background
The DataTable component wasn't originally published with Uno support due to some issues. Recently, these issues have been fixed from the Uno side, which should allow us to enable the various Uno targets for the DataTable component. 

Since Uno has fixed the issues from their side, as of the latest Uno update in [e366657a924a745c6fb4162dc4c58d2e07fc0613](https://github.com/CommunityToolkit/Tooling-Windows-Submodule/commit/e366657a924a745c6fb4162dc4c58d2e07fc0613), DataTable should work as expected on all Uno targets.

## Problem
After enabling Uno targets, there are still lingering issues with DataTable. Examples of this can be seen by enabling wasm and building the gallery:
image
image
image

## Solution
This code works as expected on UWP and Wasdk. These issues will need to be addressed and fixed on the Uno side before this component can be considered production ready. 

This ticket can be used for investigating and tracking fixes for these issues.

michael-hawker:
Fixed #559
Tested on Uno WASM, works great, just StickyHeaderBehavior not working (different issue). Just delete the MultiTarget.props so we hit all targets.

### DIFF
--- a/components/DataTable/src/MultiTarget.props
+++ b/components/DataTable/src/MultiTarget.props
@@ -1,9 +1,0 @@
-<Project>
-    <PropertyGroup>
-        <!--
-          MultiTarget is a custom property that indicates which target a project is designed to be built for / run on.
-          Used to create project references, generate solution files, enable/disable TargetFrameworks, and build nuget packages.
-        -->
-        <MultiTarget>uwp;wasdk;</MultiTarget>
-    </PropertyGroup>
-</Project>


michael-hawker:
Testspace passed, but guess the connection didn't mark it as such for GitHub or something. hmm

Arlodotexe (approved):

PR #801, opened 2026-05-12
Update slngen dependency

vgromfeld:
## What
The `GenerateAllSolution.bat` script is failing on ARM64 devices with the error described in microsoft/slngen#676. 
The `slngen` tool has been fixed, but the project is still pulling an old build of the tool.

We can manually update the tools using `dotnet tool update --all`, but it will be simpler to directly pull a fixed build 😊.

## How
I'm updating the requested version of the `slngen` tool to the latest one available: 12.0.32.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -15,7 +15,7 @@
       ]
     },
     "microsoft.visualstudio.slngen.tool": {
-      "version": "12.0.13",
+      "version": "12.0.32",
       "commands": [
         "slngen"
       ]


Arlodotexe (approved):
I'm lacking an ARM64 device to test this myself, but I reached out in the [Windows App Community](https://windowsappcommunity.com/) [Discord](https://windowsappcommunity.com/discord) via [this thread](https://discord.com/channels/372137812037730304/1509276196242849843) and a community member was kind enough to test this for us on a real ARM64 device.

We've verified that this branch is able to run slngen and open the generated solution in Visual Studio, though we haven't verified that the original bug was present on `main`.

Given that this PR is just a version bump to slngen, I only need to trust the validity of the original bug report to safely close this. Thanks for the report and fix! Approved and merged 🙂

PR #802, opened 2026-05-12
Remove duplicated sort in the measure pass

vgromfeld:
## What
The `RibbonPanel` child items are sorted twice during the measure pass.
This is not needed since the list is already sorted at the beginning of the measure.

## How
I'm removing the duplicated sorting. 
I'm also simplifying the size matching logic for the groups with `RequestedWidths`

### DIFF
--- a/components/Ribbon/src/RibbonPanel.cs
+++ b/components/Ribbon/src/RibbonPanel.cs
@@ -17,13 +17,16 @@ internal sealed partial class RibbonPanel : Panel
         // We try to limit the layout changes if the parent scrollviewer is sending values with small changes.
         availableSize.Width = Math.Floor(availableSize.Width);
 
-        var childrenByPriority = Children.OrderBy(c => c is RibbonCollapsibleGroup collapsibleGroup ? collapsibleGroup.Priority : 0);
+        var childrenByPriority = Children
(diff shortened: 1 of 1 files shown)


components/Ribbon/src/RibbonPanel.cs
```diff
@@ -17,23 +17,25 @@ protected override Size MeasureOverride(Size availableSize)
...
+            .OrderBy(c => c is RibbonCollapsibleGroup collapsibleGroup ? collapsibleGroup.Priority : 0)
+            .ToList();
+
         var desiredSize = new Size();
         foreach (var child in childrenByPriority)
         {
             var collapsibleGroup = child as RibbonCollapsibleGroup;
             var requestedWidths = collapsibleGroup?.RequestedWidths;
-            if (requestedWidths is null || collapsibleGroup?.State == Visibility.Collapsed)
+            if (requestedWidths is null || collapsibleGroup!.State == Visibility.Collapsed)
```
Avid29:
```suggestion
            if (requestedWidths is null || collapsibleGroup?.State == Visibility.Collapsed)
```

Null suppression is against the toolkit's style guide, even in situations like this where it's guaranteed not null. The original nullable comparison is fine.

Avid29 (changes requested):
Nice change, always appreciate a good optimization.

Just one comment to sort out from me

PR #804, opened 2026-05-13
Add options flyout

vgromfeld:
## What
I'm adding an optional `OptionsFlyout` button to the `Ribbon` control. It allows us to add some configuration options to the ribbon like "always show", "auto hide"...

## How
I'm adding a new set of properties to the `Ribbon` control to attached a `Flyout` to the new options button.
The options button will only be displayed if there is a flyout attached to the ribbon. It can have an access key and an automation name.

## Screenshots
| State | Screenshot |
|--|--|
| Closed | image |
| Opened | image |

### DIFF
--- a/components/Ribbon/src/Ribbon.cs
+++ b/components/Ribbon/src/Ribbon.cs
@@ -19,6 +19,8 @@ namespace CommunityToolkit.WinUI.Controls;
 [TemplateVisualState(GroupName = ScrollButtonGroupNameTemplatePart, Name = DecrementButtonStateTemplatePart)]
 [TemplateVisualState(GroupName = ScrollButtonGroupNameTemplatePart, Name = IncrementButtonStateTemplatePart)]
 [TemplateVisualState(GroupName = ScrollButtonGroupNameTemplatePart, Name = BothButtonsStateTemplatePart)]
+[TemplateVisualState(GroupName = OptionsGroupNameTemplatePart, Name = OptionsVisibleStateTemplatePart)]
+[TemplateVisualState(GroupName = OptionsGroupNameTemplatePart, Name = OptionsCollapsedStateTemplatePart)]
 public sealed partial class Ribbon : Control
 {
     private const string PanelTemplatePart = "Panel";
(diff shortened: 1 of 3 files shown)


components/Ribbon/src/Ribbon.cs
```diff
@@ -55,6 +60,60 @@ public double ScrollStep
...
+    /// The DP to store the <see cref="OptionsAccessKey"/> property value.
+    /// </summary>
+    public static readonly DependencyProperty OptionsAccessKeyProperty = DependencyProperty.Register(
+        nameof(OptionsAccessKey),
+        typeof(string),
+        typeof(Ribbon),
+        new PropertyMetadata(string.Empty));
+
+    /// <summary>
+    /// The access key to set on the options flyout button.
```
Avid29:
```suggestion
    /// Gets or sets the access key to set on the options flyout button.
```

components/Ribbon/src/Ribbon.cs
```diff
@@ -55,6 +60,60 @@ public double ScrollStep
...
+    /// The DP to store the <see cref="OptionsAccessibleName"/> property value.
+    /// </summary>
+    public static readonly DependencyProperty OptionsAccessibleNameProperty = DependencyProperty.Register(
+        nameof(OptionsAccessibleName),
+        typeof(string),
+        typeof(Ribbon),
+        new PropertyMetadata(string.Empty));
+
+    /// <summary>
+    /// The accessible name for the options button.
```
Avid29:
```suggestion
    /// Gets or sets the accessible name for the options button.
```

This is the standard comment pattern for a property

components/Ribbon/src/Ribbon.cs
```diff
@@ -55,6 +60,60 @@ public double ScrollStep
...
+    /// The DP to store the <see cref="OptionsFlyout"/> property value.
+    /// </summary>
+    public static readonly DependencyProperty OptionsFlyoutProperty = DependencyProperty.Register(
+        nameof(OptionsFlyout),
+        typeof(FlyoutBase),
+        typeof(Ribbon),
+        new PropertyMetadata(null, OnOptionsFlyoutPropertyChanged));
+
+    /// <summary>
+    /// The flyout to display when the user clicks on the ribbon options button.
```
Avid29:
```suggestion
    /// Gets or sets the flyout to display when the user clicks on the ribbon options button.
```

Avid29 (approved):
Looks good to me.

Might suggest reordering the members so the dependency properties are grouped and their corresponding implementation properties are beneath the constructor. Currently we have no official pattern.

